PR #16, opened 2015-05-08
Cache fetching

forslund:
I had some problems with the latest version, probably due to changes at geocaching.com

Earthcache threw an exception '137 not a valid cache type'. If I removed the old "earthcache" that threw the same exception so both are in there now. Tested against http://www.geocaching.com/geocache/GC5JQ55_granit

Abandoned Structure attribute was not detected for http://www.geocaching.com/geocache/GC4QGMZ_punkknackarkalendern-4-7-tribute-to-sid-vicious

It seems like both lower case image names and camel case images exists right now. To support both all attribute names are made lower case in this pull request.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -86,9 +86,9 @@ class Cache(object):
         "teamwork": "Teamwork Required",
         "thorn": "Thorns",
         "ticks": "Ticks",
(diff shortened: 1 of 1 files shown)


forslund:
Not sure what is failing, I can run all tests without issues locally...

PR #19, opened 2015-06-01
Trackable support

forslund:
Basic support for trackables (travelbugs, coins, etc.) added. This pull request provides
* Trackable class
* trackable loading in geocaching module
* tests for trackable class

Next step is extending the Cache class to include a list of trackables in the cache.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,6 +20,7 @@ Features
    -  quick loading (loads just basic info very quickly)
    -  lazy loading (create cache object and load info on demand)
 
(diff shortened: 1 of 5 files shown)


tomasbedrich:
Thank you Åke, great work!

Could you also please update README with some examples of usage?

forslund:
Updated with a short simple example. Hope it's enough, I'll add the more advanced use cases when I add support for listing trackables in caches.

PR #21, opened 2015-06-23
Quickloaded trackables

forslund:
This is a bit of a cleanup making the trackable handling a bit more inline with the rest of the pycaching module and reduces clutter in the Cache class. Now when loading the cache the trackable page is parsed and trackables are quick-loaded.
(Previously trackables were not loaded at all when parsing a cache, instead a link to the trackable search page for the cache was stored and loaded together with all trackables when the trackable property was requested.)

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -563,7 +563,7 @@ class Geocaching(object):
         logging.info("Loading details about %s...", tid)
 
         url = self._urls["trackable_details"].format(tid=tid)
-        return self.load_trackable_by_url(self, url, destination)
+        return self.load_trackable_by_url(url, destination)
 
     def get_logged_user(self, login_page=None):
         """Returns the name of curently logged user or None, if no user is logged in."""
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -585,9 +585,12 @@ class Geocaching(object):
             raise Error("Cannot load cache details page.") from e
 
         trackable_table = root.find_all("table")[1]
-        links_raw = trackable_table.find_all("a")
-        urls = [l.get("href") for l in links_raw if "track" in l.get("href")]
-        names = [re.split("[\<\>]", str(l))[2] for l in links_raw if "track" in l.get("href")]
+        urls_raw = trackable_table.find_all("a")
+        # filter out all urls for trackables
+        urls = [url.get("href") for url in urls_raw if "track" in url.get("href")]
+        # find the names matching the trackble urls
+        names = [re.split("[\<\>]", str(url))[2] for url in urls_raw if "track" in url.get("href")]
+        # create trackables and build list to return
         trackables = []
         for n, u in zip(names, urls):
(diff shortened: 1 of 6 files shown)


pycaching/trackable.py
```diff
@@ -17,7 +17,10 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
         except AttributeError:
             logging.debug("Lazy loading: %s", func.__name__)
-            self.geocaching.load_trackable(self.tid, self)
+            try:
+                self.geocaching.load_trackable_by_url(self.trackable_page, self)
+            except AttributeError:
+                self.geocaching.load_trackable(self.tid, self)
```
tomasbedrich: Wouldn't this cause an endless loop (recursion) in case of lazy loading `tid` of Trackable, which doesn't have its `trackable_page` specified? If yes, could you please handle it (some meaningful error message)?

pycaching/geocaching.py
```diff
@@ -584,4 +587,8 @@ def load_trackable_list(self, url):
         trackable_table = root.find_all("table")[1]
         links_raw = trackable_table.find_all("a")
         urls = [l.get("href") for l in links_raw if "track" in l.get("href")]
-        return [self.load_trackable_by_url(t) for t in urls]
+        names = [re.split("[\<\>]", str(l))[2] for l in links_raw if "track" in l.get("href")]
```
tomasbedrich: Could you please rename variables / add a comment / do somehow this line more readable?

tomasbedrich:
Great job @forslund, but one more thing: I would like to keep a [test coverage](https://coveralls.io/builds/2885529) above 90% (ideally above 95%). Would you please take care of it?

forslund:
I'll get right on all that!

pycaching/trackable.py
forslund: Regarding this, I see two possible solutions...
1. Catch the infinite recursion exception and raise a ValueError
2. Using has_attr(self, '_tid') to determine if the _tid property exist and if that doesn't exist raise ValueError

Personally I think I would prefer 2 since it feels bad allowing an endless recursion but using hasattr() seems to be considered unpythonic and thus 1 is to be preferred?

What's you opinion?

tomasbedrich: I would prefer slightly modified 2nd way. Just add simple condition determining if `trackable_page` or `tid` is set and call appropriate method. If neither, raise a `pycaching.errors.LoadError`.

forslund:
I Think the tests should be alright now...is there any way to get the test-settings from coveralls so I know I run with the same settings? I used `nosetests --with-coverage --cover-package=pycaching test`

Clearing up the beautiful soup parsing in `load_trackable_list()` is a bit tricky but it should be somewhat more clear now.

tomasbedrich:
You can find the exact command in `.travis.yml`. Currently it is: `nosetests --with-coverage --cover-package=pycaching` (exactly same as you used).

Thank you for your contribution! I am looking forward to more PRs. :+1:

PR #33, opened 2015-08-08
Prevent exception when location_url has no value

mrvdb:
Make sure location_url has an (empty) value, to prevent an exception 
(tested with TB6JYV5)


tomasbedrich:
Thank you again and sorry for the delay - I was on the holiday.

mrvdb:
No problem, thanks for merging.

PR #39, opened 2015-09-07
"Attended" events are considered found

### LINKED ISSUE #35: "Found it" for events
Event-caches  are marked as "not found", because the corresponding text is "Attended" and not "Found it!"

-- Uwe Zimmermann

forslund:
.found property is set accordingly. This resolves #35.

Not quite sure how to write the integration test for this. I could pickle the page I've been testing against and monkey patch the Geocache._request()-method but I'm not sure this is a great idea...


tomasbedrich:
I think it is okay to leave this without test. :)

PR #40, opened 2015-09-18
Post log

tomasbedrich:


README.rst
```diff
@@ -94,6 +96,24 @@ Using lazy loading:
 The difference is, that ``Cache`` object is created immediately and the
 page is loaded when needed (accessing the name).
 
+Post a new log to a cache:
```
tomasbedrich: Could you please make this a heading? Using: `~~~`

pycaching/cache.py
```diff
@@ -412,6 +424,7 @@ def load(self):
         else:
             trackable_page_url = None
 
+        log_page_url = root.findAll("div", "CacheDetailNavigation NoPrint")[0].find("a", {'id' : 'ctl00_ContentBody_GeoNav_logButton'})["href"]
```
tomasbedrich: Could you select this element by using its ID? The ID should be unique in the whole page, so the selector could be shortened to: `root.find(id="ctl00_ContentBody_GeoNav_logButton")["href"]`

forslund: Thanks, this is so much better!

forslund:
When we feel the this is ready for merge maybe I should do some rebasing and try to make a cleaner pull request?

tomasbedrich:
I think it is ready for merge now. Do you want me to do it, or do you plan to add more features?

forslund:
I don't plan to add more features right now. When I get hold of a travelbug I'll see if I can add logging of those but I think it's better to do that as a separate pull-request. Do you want me to clean up the pull request before merge or is it OK with all the one-line fix-commits in here?

tomasbedrich:
I think it is OK, thank you!

PR #42, opened 2015-09-26
Fix #41: Some input field don't have value

### LINKED ISSUE #41: Some input field don't have value
While posting log, some input don't have a value field.
Ex: `ctl00$ContentBody$LogBookPanel1$uxTrackables$hdnSelectedActions`

kumy:


tomasbedrich:
Thank you! Would you also be so kind and write some tests for log posting?

PR #44, opened 2015-10-20
Post log tests

forslund:
This adds some tests related to post_log().

09e0409 is simple enough, just tests the property ```log_page_url```
e7a19f4 verifies that the log text field is not an empty string. The small change in ```pycaching/cache.py``` should be added even if the test is a bit clumsy.
8f2adef is a test for the log-type matching code and might be overkill since it requires some restructuring, mocks out a lot and is generally messy.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -566,7 +566,7 @@ class Cache(object):
         return (valid_types, hidden_elements)
 
     def post_log(self, l):
-        if len(l.text) == 0:
+        if not l.text:
             raise errors.ValueError("Log text is empty")
 
         valid_types, hidden = self._load_log_page()
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -556,6 +556,9 @@ class Cache(object):
             yield t
 
     def post_log(self, l):
+        if len(l.text) == 0:
+            raise errors.ValueError("Log text is empty")
+
         log_page = self.geocaching._request(self.log_page_url)
         # Find all valid log types for the cache
         type_options = log_page.find_all("option")
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -555,20 +555,24 @@ class Cache(object):
             self.trackables.append(t)
             yield t
 
-    def post_log(self, l):
-        if len(l.text) == 0:
-            raise errors.ValueError("Log text is empty")
-
+    def _load_log_page(self):
         log_page = self.geocaching._request(self.log_page_url)
         # Find all valid log types for the cache
         type_options = log_page.find_all("option")
         valid_types = {o.get_text().lower(): o["value"] for o in type_options}
-        if l.type.value not in valid_types:
-            raise errors.ValueError("The Cache does not accept this type of log")
 
         # Find all static data fields needed for log
         hidden_elements = log_page.find_all("input", type=["hidden", "submit"])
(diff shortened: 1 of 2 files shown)


test/test_cache.py
```diff
@@ -134,6 +136,40 @@ def test_favorites(self):
...
+        def mock_load_log_page(self):
+            return ({'temporarily disable listing': '22', 'archive': '5', 'write note': '4', "didn't find it": '3', '- select type of log -': '-1', 'update coordinates': '47', 'needs maintenance': '45', 'owner maintenance': '46'}, {})
+
+        real_request = Geocaching._request
+        Geocaching._request = mock_request
+
+        real_load_log_page = Cache._load_log_page
+        Cache._load_log_page = mock_load_log_page
+
+        with self.subTest("log type"):
```
tomasbedrich: Could you please change the message to something like "invalid log type"?

test/test_cache.py
```diff
@@ -134,6 +136,40 @@ def test_favorites(self):
...
+            l = Log(text="Test log.", visited=date.today(), type = LogType.found_it)
+            with self.assertRaises(ValueError):
+                self.c.post_log(l)
+
+            l = Log(text="Test log.", visited=date.today(), type = LogType.didnt_find_it)
```
tomasbedrich: Please add a subtest for this.

pycaching/cache.py
```diff
@@ -555,17 +555,24 @@ def load_trackables(self, limit=float("inf")):
...
-        if l.type.value not in valid_types:
-            raise errors.ValueError("The Cache does not accept this type of log")
 
         # Find all static data fields needed for log
         hidden_elements = log_page.find_all("input", type=["hidden", "submit"])
-        post = {field["name"]: (field["value"] if field.has_attr("value") else "") for field in hidden_elements}
+        return (valid_types, hidden_elements)
+
+    def post_log(self, l):
+        if len(l.text) == 0:
```
tomasbedrich: Why not to use more idiomatic: `if not l.text: ...`? :wink:

tomasbedrich:
Thank you for the tests.
You are right – the mocking of some methods is a kind of wild... Could you please take a look at [unittest.mock module](https://docs.python.org/3.5/library/unittest.mock.html#)? I think we could mock the whole Geocaching class for the purpose of this test and just assert the `_request` with some arguments is called. This module could help us to do the mocking in a more standardised way.
As a last thing – could you please add a test for `_load_log_page` method?

pycaching/cache.py
forslund: Probably because I think like a C programmer =)

I'll change it

forslund:
That's probably a good idea, I'll look into it during the weekend and can hopefully update the pull request on Monday or Tuesday.

tomasbedrich:
Ok, don't hurry.

forslund:
I didn't manage to mock the entire ```Geocaching``` class without causing havoc everywhere. The solution that I was somewhat happy with placed the new test together with the load_log()-test in ```TestMethods``` and only mocks the methods ```_request``` and ```_load_log_page``` while running the new test case.

It could probably still be improved so more suggestions are welcome =)

tomasbedrich:
Okay, thank you, this solution looks good enough.. :ok_hand: I have just slightly improved your code – added a test to specific arguments, instead of just True/False call test.

PR #49, opened 2015-11-15
Credentials in file

filipek92:
Added possibility to login by credentials in file .gc_credentials

File must be in json with field username ans password

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -4,7 +4,9 @@ import logging
 import datetime
 import requests
 import bs4
+import json
 from urllib.parse import urljoin
+from os import path
 from pycaching.cache import Cache, Type, Size
 from pycaching.log import Log, Type as LogType
 from pycaching.geo import Point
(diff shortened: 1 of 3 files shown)


pycaching/geocaching.py
```diff
@@ -48,10 +50,31 @@ def _request(self, url, *, expect="soup", method="GET", login_check=True, **kwar
         except requests.exceptions.RequestException as e:
             raise Error("Cannot load page: {}".format(url)) from e
 
-    def login(self, username, password):
+    def login(self, username=None, password=None):
         """Logs the user in.
 
         Downloads the relevant cookies to keep the user logged in."""
+        if username is None or password is None:
```
tomasbedrich: Please use `not username or not pass`

pycaching/geocaching.py
```diff
@@ -48,10 +50,31 @@ def _request(self, url, *, expect="soup", method="GET", login_check=True, **kwar
...
+            try:
+                credentials_file = ".gc_credentials"
```
tomasbedrich: Please create class-level private constant, like `base_url`

pycaching/geocaching.py
filipek92: Empty password or username mean use file?

test/test_geocaching.py
```diff
@@ -105,6 +108,68 @@ def test_logout(self):
...
+            valid.write(json.dumps(credentials).encode())
+            valid.close()
+            self.g._credentials_file = os.path.basename(valid.name)
+            username, password = self.g._load_credentials()
+            self.assertEqual(_username, username)
+            self.assertEqual(_password, password)
+        try:
+            os.remove(valid.name)
+        except:
+            pass
```
tomasbedrich: Why these constructions? I think that this command should not fail. If it fails, it's also a test failure.
So could you please move all `os.remove` command into the subtest without try block? Or am I missing something?

pycaching/geocaching.py
tomasbedrich: I think it is not important in this case (both are invalid for sure), so I prefer simpler solution.

filipek92: An what about your tests, you are testing that login("","") must raise an exception. Is change to login("0", "0") and hope that user "0" will not have password "0" OK?

tomasbedrich: Yes, sure. The empty values was just example, so changing them to another invalid login credentials is sure OK.

PR #51, opened 2016-01-02
Calculate default zoom level in quick search

mkouhia:
Calculate longitude span for area, then get such zoom level, where tile width
is above this span.

Previous implementation loads unnecessarily large tiles, if area is located at
the border of two large tiles.  Hence, download more smaller tiles (typically
1, 2 or 4).


tomasbedrich:
Good job mkouhia, thank you! I have to admit that geodetic computation are not my strongpoint. :)

Issue #37, opened 2015-08-26
Retrieve original coordinates for mystery caches with corrected coordinates


tomasbedrich:
The original coordinates are located in the javascript data block as "oldLatLng".

-- Uwe Zimmermann

forslund:
Do we have the GC-code for a cache that has this issue?

tomasbedrich:
It is more feature request than bug. You can edit the coordinates for every mystery and multi cache (the pencil button next to the coordinates, see attached image).

<img width="1046" alt="snimek obrazovky 2016-01-18 v 15 03 05" src="https://cloud.githubusercontent.com/assets/4974343/12393229/b03602a0-bdf4-11e5-941c-bb9feda3a39f.png">

When you load the cache using pycaching, only the new coordinates will be parsed. I think the point of Uwe's suggestion was to retrieve the original coordinates too. Currently they are not available from the pycaching point of view.

tomasbedrich:
Are you interested in solving this?

forslund:
Yes, I didn't have time looking into it during the week but expect a pull
request tonight or tomorrow morning.
Den 23 jan 2016 14:21 skrev "Tomáš Bedřich" <notifications@github.com>:

tomasbedrich:
Don't hurry, I just want to know.

PR #57, opened 2016-01-24
Mystery original coords

forslund:
The simple solution for #37. A new property ```orginal_location``` is added to the Cache-object.

As I implemented this I started thinking that this might not be the best way, maybe we should intruduce a dictionary named additional_locations which contains any additional coordinates. Since many caches have extra locations such as coordinates for parking or for the different steps in a multi-cache.

Any thoughts?

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -594,7 +594,7 @@ class Cache(object):
 
         self.location = Point.from_string(root.find(id="uxLatLon").text)
 
-        #find original location if any
+        # find original location if any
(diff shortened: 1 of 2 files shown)


tomasbedrich:
Great! The last thing remains - to fix this subtest: https://github.com/forslund/pycaching/blob/9bfa7b0be92db75120d9c5a622b7d7f5fedcbd2b/test/test_cache.py#L79

tomasbedrich:
Thank you. Can you please add a separate issue for your idea of more locations for one cache? Including some draft – what solution would be nice according to you.

PR #65, opened 2016-08-19
Catch exception if waypoint location is invalid

weinshec:
When parsing waypoints from the HTML table it is not guaranteed that the location is in proper format, e.g. hidden stages have '???' in the location field (see e.g. GC3WTTZ). This caused an unhandled exception in `Waypoint.from_html()`.

Fixing this issue by catching the `ValueError` raised by the `geopy.Point()` initializer and setting the waypoints location to `None`.


tomasbedrich:
Great! Just one more change before merging please: use `.format()` to format a logging message.

weinshec:
Thanks for merging!

Issue #62, opened 2016-07-20
Enable request for trackable routes


FriedrichFroebel:
When accessing a trackable page via web browser I am able to download the route of the specified trackable as a KML file. Could be very helpful to enable an automated download of these routes (saving them as files would be enough from my point of view).
Seems like the download link could be parsed using `id="ctl00_ContentBody_lnkGoogleKML"`.

tomasbedrich:
Okay, I think we can do it this way:

- [ ] parse the download link you suggested in the `Trackable.load` method, store it as an instance attribute
- [ ] create `Trackable.get_KML` method which would use this attribute to make a HTTP request and return a KML as a `str`
- [ ] write tests and docs

Would you consider helping with that, please?

FriedrichFroebel:
I will have a look at this as soon as possible.
Unfortunately I could not get the liner working - it raises some `IndexError: list index out of range`.

tomasbedrich:
Can you please create a separate issue for that? It would be best, if you could include a full error message.

tomasbedrich:
Hello @FriedrichFroebel, what do you think – are you going to solve this issue? It is the last release blocker for version 3.6.

FriedrichFroebel:
I am currently on vacation, so this would probably take some days until
I will have a look at this.

tomasbedrich:
Okay, have a nice vacation! 😉  I will wait until the 10th of September, so please inform me about the progress as you get back.

FriedrichFroebel:
Just found some time to have a look at this. I currently get stuck when trying to test my current implementation as I get `AttributeError: 'Trackable' object has no attribute 'get_KML()'` after running the setup script after my changes and then trying to use some small script:
```
import pycaching
g = pycaching.login(...)  # credentials removed
trackable = g.get_trackable("TB3ZGT2")
print(trackable.name)  # works
trackable.get_KML()  # throws error
```

tomasbedrich:
I think the problem is the setup script you run again. You should only need to setup your enviroment once. If you need to test new features, either write new tests or use a test script named `test.py` inside the root of the project, which is ignored by GIT.

FriedrichFroebel:
This way the problem has gone away.

Another problem: The BeautifulSoup4 module changes the XML tags of the KML file to lowercase which seems to be a problem for Google Earth when trying to import it there.

tomasbedrich:
[...]

PR #70, opened 2016-09-06
KML routes of trackables (#62)

FriedrichFroebel:


test/test_trackable.py
```diff
@@ -124,3 +124,11 @@ def test_post_log(self, mock_request, mock_load_log_page):
                 "ctl00$ContentBody$LogBookPanel1$uxLogInfo": test_log_text,
             }
             mock_request.assert_called_with(self.t._log_page_url, method="POST", data=expected_post_data)
+
+    def test_load_KML(self):
```
tomasbedrich: Please rename the test method according to the tested method. In this case `test_get_KML()`.

pycaching/trackable.py
```diff
@@ -179,6 +199,7 @@ def load(self):
         self.owner = root.find(id="ctl00_ContentBody_BugDetails_BugOwner").text
         self.goal = root.find(id="TrackableGoal").text
         self.description = root.find(id="TrackableDetails").text
+        self.kmlurl = root.find(id="ctl00_ContentBody_lnkGoogleKML").get("href")
```
tomasbedrich: Can you please make 2 more changes?

1. rename the `kmlurl` to `kml_url` according to [PEP8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables):

   > ### Method Names and Instance Variables
   > Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.

1. Make the variable "non-public" (as defined in PEP8) – add one leading underscore – so the final name will be `_kml_url`. Then you can delete the new `@property` and `@kmlurl.setter` methods and handle the potentially missing URL in the `get_KML()` method the same way like [in the `_load_log_page()` method](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/trackable.py#L198-L199).

  The goal of this change is to hide this variable from `Trackable` public API.

test/test_trackable.py
```diff
@@ -124,3 +124,11 @@ def test_post_log(self, mock_request, mock_load_log_page):
                 "ctl00$ContentBody$LogBookPanel1$uxLogInfo": test_log_text,
             }
             mock_request.assert_called_with(self.t._log_page_url, method="POST", data=expected_post_data)
+
+    def test_get_KML(self):
+        self.kml = self.t.get_KML()
```
tomasbedrich: No need to store the variable to `self`, you can simply use local variable.

tomasbedrich:
Thanks for your great work! Just a few more changes and we are ready to go.

Issue #29, opened 2015-07-22
Parse cache data from different page


tomasbedrich:
Use this URL: http://www.geocaching.com/seek/cdpf.aspx?guid=182a3463-e46e-4401-8697-3ad3ac2a1a42&lc=10 to parse geocache data (possible 2 times speedup).

weinshec:
Do you have any clue how to retrieve the `guid` of a cache in the first place without loading the usual details page?

FriedrichFroebel:
Using the log page provides a link to the listing using the `guid` of the cache: https://www.geocaching.com/seek/log.aspx?wp=GC3RPVZ

tomasbedrich:
It is also possible to fetch the GUID using the `load_quick()` method:

```
GET https://tiles01.geocaching.com/map.details?i=GC4HTZW
{  
   "status":"success",
   "data":[  
      {  
         "name":"Zmijovec (Amorphophallus)",
         "gc":"GC3RPVZ",
         "g":"182a3463-e46e-4401-8697-3ad3ac2a1a42",
         "available":true,
         "archived":false,
         "subrOnly":false,
         "li":false,
         "fp":"115",
         "difficulty":{  
            "text":4.5,
            "value":"4_5"
         },
         "terrain":{  
            "text":1.0,
            "value":"1"
         },
         "hidden":"11/18/2012",
         "container":{  
            "text":"Regular",
            "value":"regular.gif"
         },
         "type":{  
            "text":"Traditional Cache",
            "value":2
         },
         "owner":{  
            "text":"Lindbergh007",
            "value":"850af23d-fc83-4d3c-b93a-9e5d6ae359c9"
         }
      }
   ]
}
```

The question is, if it will be faster to make two lightweight requests or one heavy.

I would suggest adding a GUID parsing to the `load_quick()` and creating a temporary `load_by_guid()` method, which would populate some basic Cache info using the print page. Then we could measure what is faster.

weinshec:
That sounds reasonable to me. I would like to give it try and will report about the performance comparison as soon as I have a first implementation.

weinshec:
So here are some numbers... I used the [timeit](https://docs.python.org/3.5/library/timeit.html) module to profile the call to `Cache.load()` and compared it to a call of `Cache.load_quick()` parsing the guid and subsequently calling `Cache.load_by_guid()` requesting the print-page. The result is based on the mean time averaging over 100 calls each.

[...]

PR #73, opened 2016-10-26
Issue29 cache loading by GUID

weinshec:
Implement cache details loading by GUID according to issue #29 
- Add method to retrieve a caches GUID
- Add method to load all cache details from its _print-page_ (with a few exceptions of properties that are not listed there)
- Add `id_` argument to `Waypoints.from_html()` allowing to specify the html-id of the waypoint table to parse

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -700,14 +700,14 @@ class Cache(object):
             self._load_guid()
         res = self.geocaching._request(self._urls["print_page"],
                                        params={"guid": self.guid})
-        if res.find("p", {"class": "Warning"}) is not None:
+        if res.find("p", "Warning") is not None:
             raise errors.PMOnlyException()
         content = res.find(id="Content")
 
         self.name = content.find("h2").text
 
         self.location = Point.from_string(
-            content.find("p", {"class": "LatLong Meta"}).text)
+            content.find("p", "LatLong Meta").text)
 
         type_img = os.path.basename(content.find("img").get("src"))
         self.type = Type.from_filename(os.path.splitext(type_img)[0])
@@ -715,8 +715,7 @@ class Cache(object):
         size_img = content.find("img", src=re.compile("\/icons\/container\/"))
         self.size = Size.from_string(size_img.get("alt").split(": ")[1])
 
-        D_and_T_img = content.find(
-            "p", {"class": "Meta DiffTerr"}).find_all("img")
+        D_and_T_img = content.find("p", "Meta DiffTerr").find_all("img")
         self.difficulty, self.terrain = [
             float(img.get("alt").split()[0]) for img in D_and_T_img
         ]
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -724,8 +724,8 @@ class Cache(object):
         self.author = content.find(
             "p", text=re.compile("Placed by:")).text.split("\r\n")[2].strip()
 
-        self.hidden = content.find(
-            "p", text=re.compile("Placed Date:")).text.split(": ")[1]
+        hidden_p = content.find("p", text=re.compile("Placed Date:"))
+        self.hidden = hidden_p.text.replace("Placed Date:", "").strip()
 
         attr_img = content.find_all("img", src=re.compile("\/attributes\/"))
         attributes_raw = [
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -729,7 +729,7 @@ class Cache(object):
 
         attr_img = content.find_all("img", src=re.compile("\/attributes\/"))
         attributes_raw = [
-            _.get("src").split("/")[-1].rsplit("-", 1) for _ in attr_img
+            os.path.basename(_.get("src")).rsplit("-", 1) for _ in attr_img
         ]
         self.attributes = {
             name: appendix.startswith("yes") for name, appendix
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -704,7 +704,7 @@ class Cache(object):
             raise errors.PMOnlyException()
         content = res.find(id="Content")
 
-        self.name = content.find("h2").get_text()
+        self.name = content.find("h2").text
 
         self.location = Point.from_string(
             content.find("p", {"class": "LatLong Meta"}).text)
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -180,6 +180,22 @@ class Cache(object):
         self._wp = wp
 
     @property
+    def guid(self):
+        """The cache GUID. An identifier used at some places on geoaching.com
+
+        :type: :class:`str`
+        """
+        return self._guid
+
+    @guid.setter
+    def guid(self, guid):
+        guid = guid.strip()
+        guid_regex = "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+        if not re.match(guid_regex, guid):
+            raise errors.ValueError("GUID not well formatted: {}".format(guid))
+        self._guid = guid
+
+    @property
     def geocaching(self):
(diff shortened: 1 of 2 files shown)


pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+            "h2", text="Long Description").find_next("div").text
+
+        self.hint = rot13(content.find(id="uxDecryptedHint").text)
+
+        self.favorites = content.find(
+            "strong", text=re.compile("Favorites:")).parent.text.split()[-1]
+
+        self.waypoints = Waypoint.from_html(content, "Waypoints")
+
+    def _load_guid(self):
```
tomasbedrich: Can you please drop this method and use `load_quick` instead of it?

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+        res = self.geocaching._request(self._urls["tiles_server"],
+                                       params={"i": self.wp},
+                                       expect="json")
+
+        if res["status"] == "failed" or len(res["data"]) != 1:
+            msg = res["msg"] if "msg" in res else "Unknown error (probably not existing cache)"
+            raise errors.LoadError("Cache {} cannot be loaded: {}".format(self, msg))
+
+        self.guid = res["data"][0]["g"]
```
tomasbedrich: If you just move this line to `load_quick`, it will have the same effect in the end. 😉

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+
+          + original_location
+          + state
+          + found
+          + pm_only
+
+        :raise .PMOnlyException: If the PM only warning is shown on the page
+        """
+        if not hasattr(self, "guid"):
+            self._load_guid()
```
tomasbedrich: Change here – please see the note for line 752 + please add a comment on this line, why using `load_quick`.

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+        res = self.geocaching._request(self._urls["print_page"],
+                                       params={"guid": self.guid})
+        if res.find("p", {"class": "Warning"}) is not None:
```
tomasbedrich: Please be consistent with the rest of the library and use `find("p", "Warning")` for finding an element by a class. The same note applies multiple times below.

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+            raise errors.PMOnlyException()
+        content = res.find(id="Content")
+
+        self.name = content.find("h2").get_text()
```
tomasbedrich: Please be consistent and use `.text` attribute.

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+
+        self.location = Point.from_string(
+            content.find("p", {"class": "LatLong Meta"}).text)
+
+        type_img = os.path.basename(content.find("h2").img.get("src"))
```
tomasbedrich: Please be consistent and use `["src"]` for HTML attribute access. Also please use `find("img")` for traversing the DOM tree.

BTW, good point with the `os.path.basename`! Can you please create a separate PR for cleaning up `` parsing like this?

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+            "p", {"class": "Meta DiffTerr"}).find_all("img")
+        self.difficulty, self.terrain = [
+            float(img.get("alt").split()[0]) for img in D_and_T_img
+        ]
+
+        self.author = content.find(
+            "p", text=re.compile("Placed by:")).text.split("\r\n")[2].strip()
+
+        self.hidden = content.find(
+            "p", text=re.compile("Placed Date:")).text.split(": ")[1]
```
tomasbedrich: Please make this a little bit more solid. This can easily break, if some user sets a date format containing `: `.

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+
+        attr_img = content.find_all("img", src=re.compile("\/attributes\/"))
+        attributes_raw = [
+            _.get("src").split("/")[-1].rsplit("-", 1) for _ in attr_img
```
tomasbedrich: You can use `os.path.basepath` like you already did. 😉

pycaching/cache.py
```diff
@@ -674,14 +684,91 @@ def load_quick(self):
...
+            in attributes_raw if not appendix.startswith("blank")
+        }
+
+        self.summary = content.find(
+            "h2", text="Short Description").find_next("div").text
+
+        self.description = content.find(
+            "h2", text="Long Description").find_next("div").text
+
+        self.hint = rot13(content.find(id="uxDecryptedHint").text)
```
tomasbedrich: You can use `uxEncryptedHint` without `rot13`.

pycaching/cache.py
```diff
@@ -831,10 +918,14 @@ def __init__(self, id=None, type=None, location=None, note=None):
         self._note = note
 
     @classmethod
-    def from_html(cls, root):
-        """Return a dictionary of all waypoints found in the page representation"""
+    def from_html(cls, root, id_="ctl00_ContentBody_Waypoints"):
```
tomasbedrich: Please remove the default value of `id_` and propagate it to the method call (line 650). Also maybe rename the variables `root` and `id` somehow more meaningful...

pycaching/cache.py
weinshec: Actually the _src_ attribute of `img` tags seems to be always accessed via `get("src")` all over the code. Do you want me to change them all (even outside `load_by_guid`)?

pycaching/cache.py
weinshec: see comment above

pycaching/cache.py
weinshec: see comment below

pycaching/cache.py
tomasbedrich: The goal is to save a bandwidth. When I already do an request to `self._urls["tiles_server"]` (either by calling `load_quick()` or `_load_guid()`), I see no reason, why not to parse *all* the data got in a response. To prevent it to be messy, I wanted you to write a simple comment inside the `load_by_guid()` method.

You are right, that the loading code is somehow redundant now, but this is intentional – everyone can have a different starting info about a cache (WP or GUID) and can end up with different result, depending on his needs and loading method used.

So please either remove the method anyway / or, if you prefer, you can "wrap" the `load_quick()` method to the `_load_guid()` to have it more clear. Somehow like this:

```python
def load_by_guid(self):
    ...
    _load_guid()
    ...

def _load_guid(self):
    """COMMENT WHY"""
    return load_quick(self)
```

weinshec: Taking different starting points (WP or GUID) into account is actually a good point. So I moved GUID loading to `load_quick()` and removed the redundant `_load_guid()` then.

pycaching/cache.py
```diff
@@ -180,6 +180,22 @@ def wp(self, wp):
...
     @property
+    def guid(self):
+        """The cache GUID. An identifier used at some places on geoaching.com
+
+        :type: :class:`str`
+        """
+        return self._guid
+
+    @guid.setter
+    def guid(self, guid):
```
tomasbedrich: 👍 for validating a guid

pycaching/cache.py
```diff
@@ -697,7 +714,7 @@ def load_by_guid(self):
         :raise .PMOnlyException: If the PM only warning is shown on the page
         """
         if not hasattr(self, "guid"):
-            self._load_guid()
+            self.load_quick()
```
tomasbedrich: Please add a brief comment _why_.

weinshec: Sure :)

PR #76, opened 2017-05-17
Fix an error where caches' found attribute were not correct

ian-howell:
Possibly related to Issue29

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -234,7 +234,7 @@ class Geocaching(object):
                 c = Cache(self, wp)
                 c.type = Type.from_string(cache_details[0].strip())
                 c.name = row.find("span", "cache-name").text
-                c.found = row.find("img", title="Found It!") is not None
+                c.found = str(row.find("svg", class_="badge")).find('found') >= 0
                 c.favorites = int(row.find(attrs={"data-column": "FavoritePoint"}).text)
                 c.state = not (row.get("class") and "disabled" in row.get("class"))
                 c.pm_only = row.find("td", "pm-upsell") is not None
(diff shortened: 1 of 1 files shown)


pycaching/geocaching.py
```diff
@@ -234,7 +234,7 @@ def search(self, point, limit=float("inf")):
                 c = Cache(self, wp)
                 c.type = Type.from_string(cache_details[0].strip())
                 c.name = row.find("span", "cache-name").text
-                c.found = row.find("img", title="Found It!") is not None
+                c.found = str(row.find("svg", class_="badge")).find('found') >= 0
```
tomasbedrich: Hello @ian-howell, thanks for your first PR to pycaching. :-)

But this is quite a hack for more that one reason. :-) The `row.find("svg", class_="badge")` part returns a beautiful soup element or `None`, depending on the lookup result. So far so good. But the `str(...)` part converts it to string, even if `None` is returned. So the code does something like `"None".find('found') >= 0`, which is really not ideal, regardless a fact, that prefered way of searching for substrings is using `in` operator: `"found" in "None"`.. :-)

So please can you split this line to 2 steps? 1st – finding the element, 2nd – finding `"found"` substring *only if* the element was found previously? And please, use the double quotes to match the project's code style. ;-)

Thank you! 👍

ian-howell:
Alright, that should fix it. Please let me know if it needs to be fleshed out anymore. Thanks!

tomasbedrich:
Great! Just please remove the comment and I will merge it. ;-)

tomasbedrich:
Thank you. Looking forward to your next PR. ;-)

Issue #77, opened 2017-06-21
cache.description is not handled gracefully


Hunter275:
If there is no long description for the cache listed on the website,  cache.description returns 
```
<div class="UserSuppliedContent">
<span id="ctl00_ContentBody_LongDescription"></span>
</div>
```

This should be handled gracefully, and just return an empty string. I will put in a pull request for this when I fix it (see #78)

PR #78, opened 2017-06-21
Fixed Long Description text parsing

Hunter275:
See #77


pycaching/cache.py
```diff
@@ -624,6 +695,16 @@ def load(self):
 
         self.location = Point.from_string(root.find(id="uxLatLon").text)
 
+        self.lat = str(root.find(id="ctl00_ContentBody_Location").find("a")).split("=")[2].split("&")[0]
+
+        self.lon = str(root.find(id="ctl00_ContentBody_Location").find("a")).split("=")[3].split("&")[0]
+
+        self.id = str(root.find(id="ctl00_ContentBody_GeoNav_logButton")).split("=")[3].split("&")[0]
```
tomasbedrich: This is definitely not a good idea to parse an URL. Please take a look at [urllib.parse](https://docs.python.org/3/library/urllib.parse.html) to know how do this correctly. 😉

On the other hand – I don't see any reason, why to parse this cache ID. We already have 2 different IDs (waypoint + guid) which are used elsewhere. So why should we want another ID, which wouldn't be used anywhere?

pycaching/cache.py
```diff
@@ -624,6 +695,16 @@ def load(self):
...
+
+        self.stateprovince = str(root.find(id="ctl00_ContentBody_Location").text.split(",")[0].split(" ")[1])
```
tomasbedrich: I appreciate your effort to parse a state and a country, but the second `split(" ")[1]` is problematic. On the website, there is a text "In STATE, COUNTRY". What if the STATE part consists of multiple words (for example – see the screenshot attached to another comment)?

Notwithstanding that, I would prefer not to solve this in pycaching, but if you need detailed address info in your application, try something like [Google Geocoding API](https://stackoverflow.com/a/6355022/570503) instead. The benefit would be, that you could geocode any point, not only the cache (for example multicache waypoints, parking places, etc.).

Therefore, I would also suggest deleting this + the country parsing (including the properties and setters of course).

pycaching/cache.py
```diff
@@ -637,7 +718,7 @@ def load(self):
 
         user_content = root.find_all("div", "UserSuppliedContent")
         self.summary = user_content[0].text
-        self.description = str(user_content[1])
+        self.description = str(user_content[1].text)
```
tomasbedrich: Nice simple fix. Thank you!

pycaching/cache.py
```diff
@@ -729,6 +810,16 @@ def load_by_guid(self):
         self.location = Point.from_string(
             content.find("p", "LatLong Meta").text)
 
+        self.lat = str(content.find(id="ctl00_ContentBody_Location").find("a").find("a")).split("=")[2].split("&")[0]
```
tomasbedrich: None of the bellow can work, because the content is loaded from cache print-page ([example](https://www.geocaching.com/seek/cdpf.aspx?guid=40041302-7f06-498c-a236-0fed7b2dbf3d)), which doesn't have these elements. So I would like to kindly ask you to delete these lines, irrespectively of what has been written in other comments.

Hunter275:
Thanks for the input, I'll take what you said and go back to the drawing board.

PR #82, opened 2017-07-19
Properly convert coordinates in the Western and Southern Hemispheres

### LINKED ISSUE #81: Longitude/Latitude are being converted incorrectly when the final product should be negative.
In lines [73-76 of geo.py](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/geo.py#L73-L76), the degree portion of coordinates is converted to negative if the corresponding measure is in the Southern Hemisphere or the Western Hemisphere. On [line 78](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/geo.py#L78), the degree and minute components of the latitude and longitude are passed on to a function which converts them to decimal degree format. In [that function](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/geo.py#L19), the degree component is added to the minute component over sixty. However, when the coordinates are in the Southern Hemisphere or the Western Hemisphere, the degree component of one or both coordinates is negative, but the corresponding minute component is positive. This has the unintended effect of converting Southern or Western coordinates by the formula `–(degree – (minute/60))`, when the correct formula would be `–(degree + (minute/60))`.

I believe this problem is a two-line fix. The previously linked code block([73-76 of geo.py](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/geo.py#L73-L76)) should be changed to the following:

```
            if "S" in string:
                latDeg *= -1
                latMin *= -1
            if "W" in string:
                lonDeg *= -1
                lonMin *= -1
```

This converts both the degree and minute portions of the coordinate to negative numbers, so that they can be added to increase the absolute value of the result, as intended.

This bug seems to have been introduced [two years ago](https://github.com/tomasbedrich/pycaching/commit/9afd34b61550dbe2e90a23eefdce42855982d5bf). I suspect the author never noticed it, as he appears to live in the Czech Republic, which is in the Northern and Eastern Hemispheres. 😜

jarhill0:
Fix #81

### DIFF
--- a/pycaching/geo.py
+++ b/pycaching/geo.py
@@ -72,8 +72,10 @@ class Point(geopy.Point):
 
             if "S" in string:
                 latDeg *= -1
+                latMin *= -1
             if "W" in string:
                 lonDeg *= -1
+                lonMin *= -1
 
             return cls(to_decimal(latDeg, latMin), to_decimal(lonDeg, lonMin))
 
--- a/test/test_geo.py
+++ b/test/test_geo.py
@@ -32,7 +32,7 @@ class TestPoint(unittest.TestCase):
 
         with self.subTest("south and west"):
             self.assertEqual(Point.from_string("S 49 45.123 W 013 22.123"),
-                             Point(-48.24795, -12.63128))
+                             Point(-49.75205, -13.36872))
 
         with self.subTest("letter together"):
             self.assertEqual(Point.from_string("N49 45.123 E013 22.123"), Point(49.75205, 13.36872))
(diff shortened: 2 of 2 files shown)


tomasbedrich:
Thank you for a fix. You are right, that I never noticed this problem, because I have never used pycaching for other coordinates than N-E. :)

Do you think, it would be possible to also write tests for this case? To prevent repeating the same error again.

jarhill0:
I will write tests.

jarhill0:
I found tests that already existed, and I corrected them to what the output should be.

PR #86, opened 2017-07-20
publish_listing log type now has the correct value

### LINKED ISSUE #84: Log type '24' is not recognized
Load any cache as `c` and run the following:

```
for l in c.load_logbook():
    print('{} by {} ({}):\n\n{}'.format(l.type, l.author, l.visited, l.text))
```

All logs will output successfully, until the end, when the following exception is raised: `pycaching.errors.ValueError: Unknown log type '24'.`.

By visiting https://www.geocaching.com/images/logtypes/24.png, one can see that log type '24' is 'Publish Listing'. Can this be added?

jarhill0:
fixes #84

### DIFF
--- a/pycaching/log.py
+++ b/pycaching/log.py
@@ -106,7 +106,7 @@ class Type(enum.Enum):
     owner_maintenance = "46"
     placed_it = "14"
     post_reviewer_note = "18"
-    publish_listing = "23"
+    publish_listing = "24"
     retract = "25"
     retrieved_it = "13"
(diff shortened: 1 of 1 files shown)


jarhill0:
It means [enable listing](https://github.com/tomasbedrich/pycaching/blob/master/pycaching/log.py#L98).

tomasbedrich:
Great, thank you. I have reviewed a code from a mobile phone, so I haven't noticed this.

Issue #89, opened 2017-11-26
cache.found = user has left comment  


Wlodarski:
If a user posts a DNF (didn't not find) or a note on geocaching.com, pychaching reports that cache is if it was "found" by that user.

Wlodarski:
By replacing line 630 in cache.py by 

>  self.found = found and ("Found It!" in found.text or "Attended" in found.text) or False

the script works for me

jarhill0:
For that matter, there's no need to have a tailing `or False`. The line could be just `self.found = found and ("Found It!" in found.text or "Attended" in found.text)`.

Wlodarski:
Fix worked. Thank you.

PR #90, opened 2017-12-03
Fix cache found status, mega event type, l10n errors

### LINKED ISSUE #88: pycaching.errors.ValueError: Unknown cache size
When the geocache account is set to another language than English, any request involving seeking cache info fails.

Example of code that fails (taken from docs):

> point = Point(56.25263, 15.26738)
> 
> for cache in geocaching.search(point, limit=50):
>     if cache.type == Type.traditional:
>         print(cache.name)

Yields:
> pycaching.errors.ValueError: Unknown cache size

Works as expected with user's preference set to English.

tomasbedrich:
Should fix #88 and #89. Please test @Wlodarski and let me know, if it works.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -319,11 +319,18 @@ class Cache(object):
 
         :type: :class:`bool`
         """
-        return self._found
+        if self._found_status:
+            return self._found_status.type in (LogType.found_it, LogType.attended)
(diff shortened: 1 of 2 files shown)


Wlodarski:
Works for me. Fixed #88 and #89 

Thank you.

PR #91, opened 2018-01-21
Added some if statements to avoid breaks in the code.

jdownie:


jdownie:
Oops. There was a bug in this pull request.

Issue #98, opened 2018-08-18
pycaching.login() has stopped working


TomasRamdahl:
I recently noticed that it no longer seems possible to log in to the API using the pycaching.login-function. This is a recent development and earlier working code, now fails when attempting to log in. There are no problems with the credentials, since they work fine when signing in on the geocaching.com web-site.

Using: pycaching  3.6.4

Error meggage:
  File "C:\Python36\lib\site-packages\pycaching\__init__.py", line 18, in login
    g.login(username, password)
  File "C:\Python36\lib\site-packages\pycaching\geocaching.py", line 135, in login
    raise LoginFailedException("Cannot login to the site "
pycaching.errors.LoginFailedException: Cannot login to the site (probably wrong username or password).

kumy:
We recently had to change auth login too in c:geo app. :/ Old login page is not working anymore. See recemtly closed issues

On August 18, 2018 7:14:26 PM GMT+04:00, TomasRamdahl <notifications@github.com> wrote:
>I recently noticed that it no longer seems possible to log in to the
>API using the pycaching.login-function. This is a recent development
>and earlier working code, now fails when attempting to log in. There
>are no problems with the credentials, since they work fine when signing
>in on the geocaching.com web-site.
>
>Using: pycaching  3.6.4
>
>Error meggage:
>File "C:\Python36\lib\site-packages\pycaching\__init__.py", line 18, in
>login
>    g.login(username, password)
>File "C:\Python36\lib\site-packages\pycaching\geocaching.py", line 135,
>in login
>    raise LoginFailedException("Cannot login to the site "
>pycaching.errors.LoginFailedException: Cannot login to the site
>(probably wrong username or password).
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/tomasbedrich/pycaching/issues/98

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mrvdb:
fyi, changing the login url to the new value in geocaching.py is sufficient for a fix.

PR #99, opened 2018-08-21
Fix login url and update according cassettes

weinshec:
+ Login URL on geocaching.com changed from `account/login` to
`account/signin`
+ Update all betamax cassettes regarding login transactions
+ Fixes issue #98


tomasbedrich:
Thank you.

Issue #106, opened 2018-08-26
Python 3.7 compatibility and PEP479


weinshec:
With Python 3.7 the handling of `StopIteration` inside generators has changed. [PEP479](https://www.python.org/dev/peps/pep-0479/) which was introduced in Python 3.5 is now enforced and raises a `RuntimeError` when `StopIteration` is used within generators. This currently affects 3 methods, i.e.

- `Geocaching.search()`
- `Cache.load_trackables()`
- `Cache.load_logbook()`

which raise a non-silent deprecation warning in Python 3.6 and a `RuntimeError` in Python 3.7.

Our CI currently tests Python versions 3.4 and 3.5 only. However, either we should fix this (easy, simply replace every occurance of `raise StopIteration` by `return`, see [PEP479](https://www.python.org/dev/peps/pep-0479/)) or we should restrict our supported Python versions.

PR #107, opened 2018-08-26
Python 3.7 support

weinshec:
+ Fix issue #106 for Python 3.7 compatibility.
+ Add Python versions 3.6 and 3.7 to testing

Should we keep version 3.4 in the Travis CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
 language: python
 sudo: false
 
-python:
-  - 3.4
-  - 3.5
-  - 3.6
-  - 3.7
+matrix:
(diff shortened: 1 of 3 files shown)


weinshec:
Seems Travis CI is lacking propper support for Python 3.7 yet (see [here](https://github.com/travis-ci/travis-ci/issues/9815)), which is why this workaround relying on the `xenial` image of Travis is currently suggested to run tests against Python 3.7

tomasbedrich:
Thank you for PR. I think that keeping older Python version in CI can only be good. 👍

PR #115, opened 2019-01-19
Improve networked tests

### LINKED ISSUE #114: Recording new tests is difficult because existing tests mess with the "session"
In my experience, trying to record a cassette for a new test doesn't work a lot of the time. A lot of tests mess with the "session" and existing cookies. I think I can fix this, but it wasn't within the scope of #113, so I didn't touch it there.

jarhill0:
Fixes #114, I believe.

This changes so that each test class has its own session, `Geocaching` instance, and recorder. Hopefully this will improve the experience of recording new tests.


tomasbedrich:
thanks 👍

Issue #117, opened 2019-03-20
Function my_finds() stops at premium cache


S0S-90:
When I try to get all my found caches with the function ``my_finds()`` the generator stops at a cache that can only be seen by premium members. (I'm not a premium member, so when I found the cache, it wasn't premium but now it is.) However it would be nice to at least see the caches found before the premium cache if not also the information about the premium cache that is available.

I think the problem is in the function ``my_logs()`` of the ``Geocaching`` class where it yields the cache by calling ``yield self.get_cache(guid=guid)`` as loading by guid does not work for premium caches. I would suggest catching this exception here and do "normal loading" of the cache, i.e. by ``get_cache(wp)``.

In order to do this the GC code (wp) has to be parsed from the website. I haven't tried that yet but I wrote a quickfix where it just returns a string when the PMOnlyException is thrown. 

So the end of the ``my_logs()`` function looks like that:

<pre>try:
    yield self.get_cache(guid=guid)
except PMOnlyException:
     # better: yield self.get_cache(gccode)
    yield "PMOnlyException"
yielded += 1</pre>

If I find the time I will try to get a better solution.

S0S-90:
As far as I am concerned this issue should be solved by pull request #118.

PR #118, opened 2019-03-24
Avoid stopping function my_logs() at premium cache

S0S-90:
see #117 

I have changed the function my_logs() so that if cache can be read by guid it is read by guid and otherwise -- i. e. if the cache is only for premium members -- it reads the cache by the GC code. So my_logs() doesn't throw an error and you can get the information of the caches you have found before the PMonly cache. It is even possible to get some information on the PMonly cache by cache.load_quick().

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -13,7 +13,7 @@ from pycaching.cache import Cache, Size
 from pycaching.log import Log, Type as LogType
 from pycaching.geo import Point
 from pycaching.trackable import Trackable
-from pycaching.errors import Error, NotLoggedInException, LoginFailedException
+from pycaching.errors import Error, NotLoggedInException, LoginFailedException, PMOnlyException
 
 
 class Geocaching(object):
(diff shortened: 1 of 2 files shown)


pycaching/geocaching.py
```diff
@@ -382,6 +382,27 @@ def _cache_from_guid(self, guid):
         print_page = self._request(Cache._urls["print_page"], params={"guid": guid})
         return Cache._from_print_page(self, guid, print_page)
 
+    def _try_getting_cache_from_guid(self, guid):
+        """tries to read a geocache from GUID page
```
tomasbedrich: Please follow the same format for docstrings as the other functions. This is important because we use it to generate docs.

S0S-90:
I've tried to implement your requested changes:
- changed history with 'git rebase -i HEAD~3'
- read gccode from redirected url (makes function _get_gccode_from_guidpage unnecessary)
- use self._request with stored URL to get the redirected url
- reformated doctring

PR #123, opened 2019-06-20
Limit requests version

tomasbedrich:
To maintain compatibility with Python 3.4.

See: https://2.python-requests.org/en/master/community/updates/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ Pycaching has following requirements:
 .. code::
 
     Python>=3.4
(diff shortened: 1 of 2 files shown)


FriedrichFroebel:
Just out of curiosity: It seems like Python 3.4 is considered [EOL](https://www.python.org/downloads/release/python-3410/). Why not dropping support for it?

tomasbedrich:
@FriedrichFroebel good point, didn't know that.

PR #124, opened 2019-06-21
Correct lowercase coordinates

twlare:
Return the correct hemisphere for an input string with lowercase characters such as "N32 46.000 w111 38.555"

### DIFF
--- a/pycaching/geo.py
+++ b/pycaching/geo.py
@@ -54,14 +54,15 @@ class Point(geopy.Point):
         This method can handle various malformed formats. Example inputs are:
 
         - :code:`S 36 51.918 E 174 46.725` or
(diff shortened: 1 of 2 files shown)


twlare:
Sorry, this is my first contribution to GitHub and I didn't understand how
to correct the error.

I corrected the lint error and added a test case to test_geo.py, but I am
not sure what to do next.

Sorry, this is probably more work for you than if you had made the change
yourself!

tomasbedrich:
Acording to the message from CI: "at least two spaces before inline comment" the fix was simple - it would be sufficient to just add one space before the comment you added. :))

Thank you for the fix and test.

It is probably more work for me, but hopefully it is more useful for your skills to fix these things on your own. :)

PR #132, opened 2020-01-13
Fix date formatting

### LINKED ISSUE #130: Error raise while log a TB
Hi, as i use in my python file

```
trackable = geocaching.get_trackable(x)
print(trackable.name, trackable.goal, trackable.description, trackable.location)
log = Log(type=LogType.discovered_it, text=logtext, visited=datetime.date.today())
tracking_code = x
trackable.post_log(log, tracking_code)
```
it raises an error like:

```
Traceback (most recent call last):
  File ".../main.py", line 95, in <module>
    main()
  File ".../main.py", line 86, in main
    trackable.post_log(log, tracking_code)
  File "...\venv\lib\site-packages\pycaching\trackable.py", line 242, in post_log
    formatted_date = format_date(log.visited, date_format)
  File ...\venv\lib\site-packages\pycaching\util.py", line 91, in format_date
    return date.strftime(date_format)
ValueError: Invalid format string
```

Tested under: 
Python 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 22:39:24) [MSC v.1916 32 bit (Intel)]

i make a change to get it work in `util.py` i do:

```
def format_date(date, user_date_format):
    """Format a date according to user_date_format."""
    # parse user format
    date_format = user_date_format.lower()
    date_format = re.split("(\W+)", date_format)
    formats = {
        "dd": r'%d',
        "d": r'%-d',
        "mmm": r'%b',
        "mm": r'%m',
        "m": r'%-m',
        "yyyy": r'%Y',
        "yy": r'%y',
    }
    date_format = "".join((formats[c] if c in formats else c for c in date_format))
    if not type(date_format) == str:
        return date.strftime(date_format)
    else:
        return date_format
```

that works for me.
please excuse the long post, i am not familiar with this.

FriedrichFroebel:
This fixes some date formatting issues on some systems as reported in #130. It seems like Windows does not like the `-` flag for date formatting which is stated as platform-dependent inside the Python docs.

This should close #130.

### DIFF
--- a/pycaching/util.py
+++ b/pycaching/util.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import logging
+import platform
 import re
 import warnings
 import inspect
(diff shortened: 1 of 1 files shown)


tomasbedrich:
This seems to solve the #130. 👍

The alternative solution may be to replace `-` with `#` for Windows (as noted on https://strftime.org/). Simple detection could look like this:

```python
import platform
eat_zero_prefix = "#" if platform.system() == "Windows" else "-"
format = {
    "d": r'%{}d'.format(eat_zero_prefix),
    ...
}
```

Feel free to use whichever solution you like more.

Also, I don't know if it's possible, but we should probably add Windows to test matrix, to prevent this kind of problems.

FriedrichFroebel:
Oh, must have missed the part regarding `#`. As this would be the cleaner solution, I am going to replace my current workaround with it.

FriedrichFroebel:
I currently do not have enough time to investigate CI stuff, but it seems that Windows tests should be available in theory, as in [this example](https://www.scivision.dev/travis-ci-osx-python/).

Issue #128, opened 2019-11-20
Retrieve the number of founds/visits of a cache


Gebruikertje:
Is it possibility to add the functionality to retrieve the number of founds of cache, just like retrieving the number of favorites?

Regards, Peter

tomasbedrich:
Thanks for a suggestion. Currently I don't have time to work on this, but as discussed in email previously:
> it shouldn't be difficult to add that functionality because the number of found caches is on each page in the header.
So if any contributor (or you) would like to take care of this, it would be welcome.

FriedrichFroebel:
The statement regarding the header seems to be wrong, as this feature request is about the log count from the logbook of the corresponding cache instead of the log count by the currently logged in user.

It seems like the needed data can be retrieved from the cache details page using the following block:

    <span id="ctl00_ContentBody_lblFindCounts"><p class="LogTotals">

Inside it, the different log types seem to be enumerated using an image and the count in alternating order.

We may add a `log_count` dictionary to the `Cache` class for this using the `log.Type` as the key and the corresponding count as the value. This way we do not restrict this to the founds, but provide other log type counts as well.

tomasbedrich:
Thanks for correcting me! I like your implementation idea, looking forward to see a PR. :)

PR #135, opened 2020-02-24
Add log counts loaders

FriedrichFroebel:
This adds loaders for the log counts using the cache details and the print pages, as per #128.

At the moment this introduces some duplicate code where both pages share the implementation. I am not sure whether it makes sense to combine them into one method or move them to a `LogCount` class inside the `log` module.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -167,7 +167,7 @@ class Cache(object):
         known_kwargs = {"name", "type", "location", "original_location", "state", "found", "size",
                         "difficulty", "terrain", "author", "hidden", "attributes", "summary",
                         "description", "hint", "favorites", "pm_only", "url", "waypoints", "_logbook_token",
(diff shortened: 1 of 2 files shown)


pycaching/cache.py
```diff
@@ -845,6 +863,97 @@ def load_by_guid(self):
...
+            types.append(type)
+
+        # Prevent possible wrong assignments when the list sizes differ for some unknown reason.
+        if not len(values) == len(types):
+            raise errors.ValueError(
+                "Different list sizes getting log counts: {} types and {} counts.".format(
+                    len(types), len(values)))
+
+        # Finally create the mapping.
+        log_counts = {}
```
tomasbedrich:
```suggestion
        log_counts = dict(zip(types, value))
```
may be more readable

FriedrichFroebel: Thanks for the hint. I have changed it in both methods.

PR #142, opened 2020-04-22
Implement property 'Personal Note'

CachingFoX:
I expanded the class Cache to read the **Personal Note** or **Cache Note** from a cache listing. The note can be read from the listing and from the print form.

There is a new property in the class Cache called 'personal_note' including the setter. I add a new test case and cassette.


test/test_cache.py
```diff
@@ -205,6 +209,11 @@ def test_load(self):
                     cache = Cache(self.gc, "GC3AHDM")
                     cache.load()
 
+        with self.subTest("Personal Note"):
+            with self.recorder.use_cassette('cache_personal_note'):
+                cache = Cache(self.gc, "GC1FPN1")  # a location less cache
+                self.assertEqual("Wandern, Wandern, Wandern, ...", cache.personal_note)
```
tomasbedrich, line 215: This is a bit problematic, because the personal note will be different for everyone who is re-running this test (= re-recording a cassette).. I am not sure, how (and whether) we should test this.

CachingFoX: Yes you are right Personal Note is a PM feature. 

If I record the cassette for my test case I have the same question, what happens if some other records the cassette. D'am! I have no real idea how we can solve this problem. 

By the way, I have a similar problem for a still existing test case with compares the number of founds. After I updated the cassette the test case fails.

tomasbedrich: Let's not test it, since we don't know how.

Feel free to delete offending tests as well. But for example, testing number of founds can be re-written to `assert num_of_founds >= 100` (which presumes non-descending tendency of this number and tests correct parsing at least). So please delete only those, which really cannot be fixed.

Issue #144, opened 2020-04-22
Load of results doesn't stop


CachingFoX:
Is the limit bigger than the number of found caches, the reload of next search results are not stopped.

**Test**
```
import pycaching
from pycaching.geo import Point
from pycaching import Cache
import logging

logging.root.setLevel(logging.DEBUG)

geocaching = pycaching.login()
results = geocaching.search(Point(38.5314833, -28.63125), limit=1000)
for item in results:
    print(item)
```

**Log**
```
...
GC6M8KX
GC74M6Z
GC4T9N0
DEBUG:root:Cache parsed: GC5TBHV
DEBUG:root:Loading page from start_index 105
DEBUG:root:Using AJAX search endpoint
GC5TBHV
DEBUG:urllib3.connectionpool:https://www.geocaching.com:443 "GET /play/search/more-results?origin=38.5314833%2C+-28.63125&startIndex=105&ssvu=2&selectAll=false HTTP/1.1" 200 44
DEBUG:root:Loading page from start_index 106
DEBUG:root:Using AJAX search endpoint
DEBUG:urllib3.connectionpool:https://www.geocaching.com:443 "GET /play/search/more-results?origin=38.5314833%2C+-28.63125&startIndex=106&ssvu=2&selectAll=false HTTP/1.1" 200 44
DEBUG:root:Loading page from start_index 107
DEBUG:root:Using AJAX search endpoint
DEBUG:urllib3.connectionpool:https://www.geocaching.com:443 "GET /play/search/more-results?origin=38.5314833%2C+-28.63125&startIndex=107&ssvu=2&selectAll=false HTTP/1.1" 200 44
DEBUG:root:Loading page from start_index 108
DEBUG:root:Using AJAX search endpoint
DEBUG:urllib3.connectionpool:https://www.geocaching.com:443 "GET /play/search/more-results?origin=38.5314833%2C+-28.63125&startIndex=108&ssvu=2&selectAll=false HTTP/1.1" 200 44
DEBUG:root:Loading page from start_index 109
DEBUG:root:Using AJAX search endpoint
DEBUG:urllib3.connectionpool:https://www.geocaching.com:443 "GET /play/search/more-results?origin=38.5314833%2C+-28.63125&startIndex=109&ssvu=2&selectAll=false HTTP/1.1" 200 44
DEBUG:root:Loading page from start_index 110
DEBUG:root:Using AJAX search endpoint
[...]

PR #145, opened 2020-04-30
Issue 144 stop search

CachingFoX:
A bug fix for issue #144 

The while loop will be stopped if there no table rows elements in loaded results

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -250,6 +250,9 @@ class Geocaching(object):
                     for label in cache_sizes_filter_wrapper.find_all("label")
                 }
 
+            if len(geocaches_table.find_all("tr")) == 0:
+                return
+
             # parse caches in result
             for start_index, row in enumerate(geocaches_table.find_all("tr"), start_index):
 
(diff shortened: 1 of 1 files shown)


CachingFoX:
I check the Travis CI reports: builds for Python 3.5 and 3.7 run fine, but with Python 3.6 the build raise an exception. I don't understand why?

FriedrichFroebel:
Wouldn't be `if not geocaches_table.find_all("tr"))` enough? And we might should save `geocaches_table.find_all("tr")` in an own variable before and use it in line 254 as well to reduce duplicate code.

CachingFoX:
> Wouldn't be `if not geocaches_table.find_all("tr"))` enough? 
Yes it could! I changed this.
> And we might should save `geocaches_table.find_all("tr")` in an own variable before and use it in line 254 as well to reduce duplicate code.
I did it :-) and push the update to Github.

Please can you check my PR, again?

FriedrichFroebel:
I added a new issue for the failing test as it seems to be unrelated to this PR, so this is ready to be merged.

Issue #134, opened 2020-01-17
Add Windows tests to CI


tomasbedrich:
To prevent regression of errors like #130.

tomasbedrich:
Related comment: https://github.com/tomasbedrich/pycaching/pull/132#issuecomment-575495643

PR #150, opened 2020-05-15
Add Windows runner configuration

FriedrichFroebel:
This adds Windows support for the Travis tests (see #134).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
 
 install:
   - python3 setup.py build install
-  - pip install python-coveralls
+  - pip install --upgrade coveralls
 
 script:
   python3 setup.py lint test
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
         - python -m pip install --upgrade pip
       install:
         - python setup.py build install
-        - pip install python-coveralls
+        - pip install --upgrade coveralls
       script:
         - python setup.py lint test
       env: PATH=/c/Python37:/c/Python37/Scripts:$PATH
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,11 +2,19 @@ language: python
...
+    - name: "Python 3.7 on Windows"
+      os: windows
+      language: shell  # `language: python` is not supported for now.
+      before_install:
+        - choco install python --version 3.7.7
+        - python -m pip install --upgrade pip
+      env: PATH=/c/Python37:/c/Python37/Scripts:$PATH
 
 install:
   - python3 setup.py build install
```
tomasbedrich, line 20: I think on Windows the `python3` command is not present by default, so this must be overridden.

FriedrichFroebel: I am currently working on it and already replaced these. Let's see if this is enough.

FriedrichFroebel:
I fixed the coverage report with `coveralls` as well, but I am not sure about the following message:

> PyYAML is not installed, skipping .coveralls.yml.

@tomasbedrich As I am not sure about it: Should we install `pyyaml` as well to eliminate this message and therefore allow handling the `.coverall.yml` of the repository as well?

tomasbedrich:
[According to docs](https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#configuration):

> If coveralls-python is being run on TravisCI, it will automatically set the token for communication with coveralls.io. Otherwise, you should set the environment variable COVERALLS_REPO_TOKEN, ...

So I think we are safe to delete `.coveralls.yml`.

Issue #75, opened 2016-12-20
UTFGrid parsing is not working


tomasbedrich:
Groundspeak changed the dimensions of UTFGrid `Block`s from 3 to 6. As a result, `Block._get_corrected_limits` doesn't work, therefore the whole `search_quick` doesn't work either.

See https://tiles01.geocaching.com/map.info?x=8800&y=5574&z=14

tomasbedrich:
@mkouhia: Based on your previous work, would you like to update the parsing code?

mkouhia:
Unfortunately I cannot allocate time to work on this. Thanks for asking,
though.


20.12.2016 4.08 ap. "Tomáš Bedřich" <notifications@github.com> kirjoitti:

@mkouhia <https://github.com/mkouhia>: Based on your previous work, would
you like to update the parsing code?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/tomasbedrich/pycaching/issues/75#issuecomment-268134277>,
or mute the thread

twlare:
Tomas, is this functionality still relevant? Would you like someone to work on it?

tomasbedrich:
It makes sense for me to fix this only if it takes less than 1-2 days. But for me it is not possible to fully understand the relevant code to fix so quickly. So if you can do better, go for it. ;)

But honestly, unless _you_ use it, I doubt that there are many people waiting for this fix.. Personally I think, that solving #29 may be more interesting.

twlare:
Hi Tomas,

You're right - I'm not using the UTFGrid function in anything I do with the
library. I was just looking at solving some of the problems.

I don't understand what's left to do with problem #29. I see that some code
has been written and tested for speed. What exactly needs to be done
besides the refactoring that you said you would do?

Tim

pkubiak:
Hi all,
nowadays there is an API endpoint used on https://www.geocaching.com/play/map/ :
```
https://www.geocaching.com/api/proxy/web/search?box=50.256919%2C18.943434%2C50.205967%2C19.027462&take=500&asc=true&skip=0&sort=distance&origin=50.2314498%2C18.985447900000054
```
it returns JSON with a lot of cache information:
```json
{
    "results": [
        {
            "id": 6358428,
            "name": "Ligota #5 Bolesław Leśmian",
            "code": "GC7A78G",
            "premiumOnly": false,
            "favoritePoints": 4,
            "geocacheType": 2,
            "containerType": 2,
[...]

PR #151, opened 2020-06-07
Feature search_rect via API

pkubiak:
I have implemented new method `Geocaching.search_rect`, which allow to search for caches in given Rectangle. It use new groundspeak API, the same as https://www.geocaching.com/play/map/.
Discussed in: https://github.com/tomasbedrich/pycaching/issues/75

Technical details:
- It requires login
- Method request caches in parts (by default 50 caches per query) in lazy way.
- During my experiment I was able to download 8600 cached in 3:40 min.
- There is an API rate limit (hit after requesting ~2000caches) signaled with `429 Too many requests` HTTP error. Now it is handled with new `TooManyRequests` exception type and appropriate wait time before continuing.
- Method return PM caches for BM, but without location information.
- Information about total number of caches in requests area is available, but not exposed via method.


pkubiak:
Should I test the API Rate Limiting logic? It requires to download ~2000 caches :)

pkubiak:
> [Coverage Status](https://coveralls.io/builds/31289875)
> 
> Coverage decreased (-0.4%) to 94.478% when pulling **[99cfe49](https://github.com/tomasbedrich/pycaching/commit/99cfe49f29ccf3b6e99fdea1a629cda7a6fb0e4b) on pkubiak:feature-search-rect-via-api** into **[c21ff39](https://github.com/tomasbedrich/pycaching/commit/c21ff39005a2b2d9dda2cafc0dc585deba60caed) on tomasbedrich:master**.

Should I test the API Rate Limiting logic? It requires to download ~2000 caches :)

pycaching/geocaching.py
```diff
@@ -356,6 +363,51 @@ def search_quick(self, area, *, strict=False, zoom=None):
...
+        """
+        Return a generator of caches in given Rectange area.
+
+        :param rect: Search area.
+        :param per_query: Number of caches requested in single query.
+        :param sortby: Order cached by given criterion.
+        :param origin: Origin point for search by distance.
+        """
+        assert sortby in {
+            'containersize', 'datelastvisited', 'difficulty', 'distance', 'favoritepoint',
```
FriedrichFroebel, line 376: I think we should use an enumeration here.

pycaching/geocaching.py
```diff
@@ -356,6 +363,51 @@ def search_quick(self, area, *, strict=False, zoom=None):
 
     # add some shortcuts ------------------------------------------------------
 
+    def search_rect(self, rect: Rectangle, *, per_query: int = 50, sortby: str = 'datelastvisited', origin=None):
+        """
+        Return a generator of caches in given Rectange area.
+
+        :param rect: Search area.
+        :param per_query: Number of caches requested in single query.
+        :param sortby: Order cached by given criterion.
```
FriedrichFroebel, line 372: Please use a consistent naming style. This should probably be `sort_by` instead.

pycaching/cache.py
```diff
@@ -153,6 +153,45 @@ def _from_print_page(cls, geocaching, guid, soup):
         cache_info["log_counts"] = Cache._get_log_counts_from_print_page(soup)
         return Cache(geocaching, **cache_info)
 
+    @classmethod
+    def _from_api_record(cls, geocaching, record):
+        """Create a cache instance from a JSON record returned by API."""
+        assert record['cacheStatus'] in {0, 1}
```
FriedrichFroebel, line 159: We should probably provide a comment about the meaning of the values here.

test/test_geo.py
```diff
@@ -144,15 +144,15 @@ def setUp(self):
 
     def test_bounding_box(self):
         bb = self.p.bounding_box
-        sw, ne = bb.corners
+        nw, se = bb.corners
```
FriedrichFroebel, line 147: What is the reason for changing this?

test/test_geocaching.py
```diff
@@ -96,6 +96,57 @@ def test_search_quick_match_load(self):
...
+                    'GC11PRW', 'GC1TYYG', 'GC7JRR5', 'GC1GW54', 'GC161KR', 'GC7KDWE', 'GC8D303'
+                ])
+
+                # Check if caches are sorted by distance to origin
+                distances = [great_circle(cache.location, origin).meters for cache in caches]
+                self.assertEqual(distances, sorted(distances))
+
+    def test_api_rate_limit(self):
+        """Test recovering from API rate limit exception."""
+        # WARNING: casset for this test has been manualy edited!
```
FriedrichFroebel, line 134: There is a typo in this comment.

And could you please elaborate on the changes here to allow re-recording and modifying the cassette if required?

pkubiak: I have simplified this testcase by removing any manual editing, cassette is quite big, but should be fully reproducible.

pycaching/cache.py
pkubiak: ok, I have defined new enum `Status`

pycaching/geocaching.py
pkubiak: fixed

pycaching/geocaching.py
pkubiak: done, I have defined enum `SortOrder`

test/test_geo.py
pkubiak: Documentation for `Rectangle` stated:
```
        """Create rectangle defined by its corners.

        :param .Point point_a: Top left corner.
        :param .Point point_b: Bottom right corner.
        """
```
But the location of `point_a` and `point_b` was not enforced anyway. And the behaviors of `Rectangle` class was fully dependent of passed arguments.

I add fix for that:
```python
        if point_a.latitude < point_b.latitude:
            point_a.latitude, point_b.latitude = point_b.latitude, point_a.latitude
        if point_a.longitude > point_b.longitude:
            point_a.longitude, point_b.longitude = point_b.longitude, point_a.longitude
```
And now `Rectangle` should be the same regardless of passed corners.

This enforcement changed the test a little.

pycaching/geocaching.py
```diff
@@ -513,3 +512,18 @@ def my_dnfs(self, limit=float('inf')):
         :param limit: The maximum number of results to return (default: infinity).
         """
         return self.my_logs(LogType.didnt_find_it, limit)
+
+
+class SortOrder(enum.Enum):
+    """Enum of possible cache sort orderings returned in Groundspeak API."""
+    # NOTE: extracted from https://www.geocaching.com/play/map/public/main.2b28b0dc1c9c10aaba66.js
+    containersize = "containersize"
```
FriedrichFroebel, line 520: I am not sure, but shouldn't we name the variables using underscores to better comply with the general code style?

test/test_geocaching.py
```diff
@@ -129,12 +129,26 @@ def test_search_rect(self):
...
+        cache_pm = self.gc._try_getting_cache_from_guid('328927c1-aa8c-4e0d-bf59-31f1ce44d990')
+        cache_pm.load_quick()  # necessary to get name for PMonly cache
+        self.assertEqual("Nidda: jenseits der Rennstrecke Reloaded", cache_pm.name)
+
+
+class TestAPIRateLimiting(NetworkedTest):
+    def test_recover_from_exceeding_rate_limit(self):
         """Test recovering from API rate limit exception."""
-        # WARNING: casset for this test has been manualy edited!
+        # WARNING: recording this test require performing a lot of requests!
```
FriedrichFroebel, line 147: Typo: "require" -> "requires".

test/test_geocaching.py
```diff
@@ -129,12 +129,26 @@ def test_search_rect(self):
...
+            cache = self.gc._try_getting_cache_from_guid('15ad3a3d-92c1-4f7c-b273-60937bcc2072')
+            self.assertEqual("Nekonecne ticho", cache.name)
+
+        # get PMonly cache from GC code (doesn't load any information)
+        cache_pm = self.gc._try_getting_cache_from_guid('328927c1-aa8c-4e0d-bf59-31f1ce44d990')
+        cache_pm.load_quick()  # necessary to get name for PMonly cache
+        self.assertEqual("Nidda: jenseits der Rennstrecke Reloaded", cache_pm.name)
+
+
+class TestAPIRateLimiting(NetworkedTest):
```
FriedrichFroebel, line 144: Shouldn't we use the usual naming scheme here, so `TestApiRateLimiting`?

pycaching/geocaching.py
```diff
@@ -363,19 +364,17 @@ def search_quick(self, area, *, strict=False, zoom=None):
 
     # add some shortcuts ------------------------------------------------------
 
-    def search_rect(self, rect: Rectangle, *, per_query: int = 50, sortby: str = 'datelastvisited', origin=None):
+    def search_rect(self, rect: Rectangle, *, per_query: int = 50, sort_by: str = 'datelastvisited', origin=None):
```
FriedrichFroebel, line 367: Why is the default parameter of `sort_by` not using the enumeration entry? Anyway, the specified type does not seem to be correct, as with the current implementation this could either be a string or an enum entry - or am I missing something?

test/test_geocaching.py
FriedrichFroebel: With the change it is stated that one has to be perform a lot of requests. I am not sure about this and how to correctly handle it.

test/test_geo.py
FriedrichFroebel: Makes sense.

test/test_geocaching.py
pkubiak: https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

pycaching/geocaching.py
pkubiak: Should I also use all upper-case for enum names?

https://docs.python.org/3/library/enum.html#creating-an-enum

test/test_geocaching.py
FriedrichFroebel: Okay, did not know this ;)

pycaching/geocaching.py
FriedrichFroebel: As far as I have seen, all our enums use lower-case names, so we should probably stick with this for now.

pycaching/geocaching.py
tomasbedrich: I suppose the goal here was to make the interface simpler for end-user. I am also for using enum type as a default, but we can allow both here by leveraging the idempotence of Enums:
```python
>>> class Foo(enum.Enum):
...   a = 1
... 
>>> Foo(1)
<Foo.a: 1>
>>> Foo(Foo.a)
<Foo.a: 1>
```

pycaching/geocaching.py
tomasbedrich: I would prefer using underscores as @FriedrichFroebel says.

Having all enums upper-case would be nice, but it's a major interface change and it's not worth it.

pycaching/geocaching.py
```diff
@@ -461,3 +512,18 @@ def my_dnfs(self, limit=float('inf')):
...
+
+
+class SortOrder(enum.Enum):
+    """Enum of possible cache sort orderings returned in Groundspeak API."""
+    # NOTE: extracted from https://www.geocaching.com/play/map/public/main.2b28b0dc1c9c10aaba66.js
+    containersize = "containersize"
+    datelastvisited = "datelastvisited"
+    difficulty = "difficulty"
+    distance = "distance"
+    favoritepoint = "favoritepoint",
```
tomasbedrich, line 524: extra comma

pycaching/geocaching.py
```diff
@@ -356,6 +364,49 @@ def search_quick(self, area, *, strict=False, zoom=None):
...
+
+        total, offset = None, 0
+        while (total is None) or (offset < total):
+            params['skip'] = offset
+
+            try:
+                resp = self._request(self._urls['api_search'], params=params, expect='json')
+            except TooManyRequests as e:
+                import time
+                time.sleep(e.rate_limit_reset + 5)
```
tomasbedrich, line 401: This is neat for most cases.

But as a pycaching user, I would like to have a control over rate-limit behavior. For example, I may want to do something else instead of sleeping and get back to downloading caches later.

There should be some parameter allowing propagation of this exception and some way of resuming the search loop. From implementation point of view, I have no preference how to solve it.

test/test_geocaching.py
FriedrichFroebel: @tomasbedrich What do you think about this specific test case?

tomasbedrich: I understand that re-recording this cassette is expensive, but I expect that we do not have to do it more often than a few times a year. Therefore, I see this solution as good enough. If it turns to be a problem, we can always switch to mocking. Is there anything else which I miss?

pycaching/geocaching.py
pkubiak: :+1:  done. I have introduced `wait_sleep` flag, which if set to `False` will result in yielding `None`s instead of sleeping. Resuming generator broken with `Exceptions` seems to be not possible.

pycaching/geocaching.py
pkubiak: fixed

test/test_geocaching.py
pkubiak: removed

pycaching/geocaching.py
pkubiak: fixed

pycaching/geocaching.py
pkubiak: :+1: fixed

test/test_geocaching.py
pkubiak: I discovered that rate limits refer to number of performed request, but not to number of downloaded caches, so I limited number of caches downloaded in each requests and cassettes are small now :) I also increased default `per_query` limit to `200`.

pycaching/geocaching.py
tomasbedrich: The parameter name is not very descriptive. :) But I like the solution from technical point of view. 👍

tomasbedrich:
Approved from my point of view. ✅ Let's just wait also for @FriedrichFroebel.

Do you want me to solve the docs, cleanup and integration as mentioned above? Or do you want to finish it yourself?

pkubiak:
> Approved from my point of view.  Let's just wait also for @FriedrichFroebel.
> 
> Do you want me to solve the docs, cleanup and integration as mentioned above? Or do you want to finish it yourself?

Thank you, for the review :) There is still one travis issue in python3.5, but I will try to fix it today. 

If it is not a problem for you, I would be grateful for solving docs, cleanup and integration :)

pycaching/geocaching.py
```diff
@@ -364,47 +380,58 @@ def search_quick(self, area, *, strict=False, zoom=None):
...
+    ):
         """
         Return a generator of caches in given Rectange area.
 
         :param rect: Search area.
-        :param per_query: Number of caches requested in single query.
+        :param int per_query: Number of caches requested in single query.
         :param sort_by: Order cached by given criterion.
-        :param origin: Origin point for search by distance.
+        :param origin: Origin point for search by distance
```
FriedrichFroebel, line 398: Is there a reason for removing the full stop?

pycaching/errors.py
```diff
@@ -57,12 +57,17 @@ class ValueError(Error, ValueError):
...
+    def __init__(self, url: str, rate_limit_reset: int = 0):
         """
         :param url: Requested url
         :param rate_limit_reset: Number of seconds to wait before rate limit reset.
         """
         self.url = url
         self.rate_limit_reset = rate_limit_reset
+
+    def wait_for(self):
+        """Wait enought time to release Rate Limits."""
```
FriedrichFroebel, line 71: Typo "enought" -> "enough".

FriedrichFroebel:
I have two minor comments, after this (and preferably fixing the failing test) it should be okay.

pycaching/errors.py
pkubiak: fixed

pycaching/geocaching.py
pkubiak: nope, fixed

pycaching/cache.py
```diff
@@ -181,7 +181,7 @@ def _from_api_record(cls, geocaching, record):
             # userDidNotFind
         )
 
-        # NOTE: postedCoordinates is not available for pm_only
+        # NOTE: Basic Members has no access to postedCoordinates of Premium-only caches
```
FriedrichFroebel, line 184: Typo "has" -> "have".

tomasbedrich:
Sorry guys, my priorities are currently elsewhere... As the situation is unlikely to change from my side and I don't want to block this anymore, I have released 4.2.0 and we can solve the docs and cleanup later.

PR #155, opened 2020-08-09
* added new date pattern

RobWei:
Recently i got this ValueError.
`pycaching.errors.ValueError: Unknown date format - '12.06.20'.`
My GC profile language is german.
So i added date format "%d.%m.%y".


RobWei:
I added the test. I hope it's done right.

FriedrichFroebel:
Unfortunately not completely: You are fixing the parser in `util.py`, but your new test is for the formatter.

In lines 17 and 18 of `test_util` (in `test_parse_date`) there should be the same patterns as in `parse_date` (they are actually out of sync already, so feel free to just copy over the the `patterns` variable or add `"%d-%m-%Y"` there as well).

FriedrichFroebel:
Fixed in #201.

PR #156, opened 2020-09-02
Add new attributes

twlare:
Add attributes `Bonus cache`, `Power trail`, `Challenge cache`, and `Geocaching.com solution checker` that were added by Groundspeak on 1-sep-2020.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -61,7 +61,7 @@ class Cache(object):
         "hike_short": "Short hike (less than 1km)",
         "hiking": "Significant hike",
         "horses": "Horses",
-        "hqsolutionchecker": "Geocaching.com solution checker",  # id = 72
+        "hqsolutionchecker": "HQ solution checker",  # id = 72
         "hunting": "Hunting",
         "jeeps": "Off-road vehicles",
         "kids": "Recommended for kids",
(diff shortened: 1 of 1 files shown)


pycaching/cache.py
```diff
@@ -59,6 +61,7 @@ class Cache(object):
         "hike_short": "Short hike (less than 1km)",
         "hiking": "Significant hike",
         "horses": "Horses",
+        "hqsolutionchecker": "Geocaching.com solution checker",  # id = 72
```
FriedrichFroebel, line 64: I am not sure about this, as it differs from the original description `HQ solution checker`. As far as I have seen, the existing attributes are using the official text from the corresponding SVG sprite (see `pycaching.util.get_possible_attributes` method), so we should probably do this here as well.

tomasbedrich:
LGTM, but let's address @FriedrichFroebel's comment.

pycaching/cache.py
twlare: I used the official text from the geocaching website, but I see what you mean about the SVG sprite.

```
  {
    "id": 72,
    "name": "Geocaching.com solution checker",
    "hasYesOption": true,
    "hasNoOption": false,
    "yesIconUrl": "https://www.geocaching.com/images/attributes/hqsolutionchecker-yes.png",
    "noIconUrl": "https://www.geocaching.com/images/attributes/hqsolutionchecker-no.png",
    "notChosenIconUrl": "https://www.geocaching.com/images/attributes/hqsolutionchecker.png"
  }
```

PR #163, opened 2021-04-24
Get logged user from JS

### LINKED ISSUE #162: Login at geocaching.com doesn't work anymore
After Groundspeak change page structure, login method raise an error. The page header are build up dynamically and  ```Geocaching.get_logged_user()``` doesn't find the user name.

A very quick and dirty hack will be
```
        def get_logged_user(self, login_page=None):
...
        logging.debug("Checking for already logged user.")
        s = login_page.find_all("script")
        for i in s:
            m = re.search(r'"username":."(.*)"', str(i))
            if m:
                return m[1]

        return None
```

tomasbedrich:
Close #162


FriedrichFroebel (changes requested):
The tests seem to fail as the cassettes have not yet been updated. We should probably do this before merging to verify the correct behaviour. Besides this, the fix looks good to me.

tomasbedrich:
You are right. I was struggling with some errors in tests, when I simply deleted the cassettes and re-recorded them. Currently, I am busy at work so if you have more time, feel free to contribute by fixing (at least some) of tests. Otherwise, I can probably fix it next weekend.

FriedrichFroebel:
I just fixed the obvious errors from the test run, which leaves one failing test. As I am currently short of time, I have not made an attempt to take a deeper look at it.

Additional notes:

* Travis states that it will shutdown tomorrow:

    >  Hi! It seems there are still plenty of builds running. We're pushing back the shutdown date to June 15th. Please consider migrating to travis-ci.com.

  We should probably migrate the CI stuff in the near future, maybe directly to GitHub Actions.
* The regex match access fix seems to be required for Python 3.5 only which already is EOL. So we might consider dropping it and reverting the change.

PR #167, opened 2021-11-14
Start migration to new pytest based test structure

### LINKED ISSUE #133: Deprecation warning running tests
Running `python3 setup.py test` produces the following output:

```
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
[...]
```

The cause seems to be that `setuptools.command.test.test` is used, which is now deprecated.

tomasbedrich:
- start migration to new pytest based test structure
- migrate build system to Flit
- migrate CI to GitHub Actions
- update CONTRIBUTING guide
- fix login tests
- re-record all cassettes
- other refactoring

Fixes #133, #162 + addresses Travis note from #163.

The core motivation is that I am not able to fix login tests for #162, because I don't understand what's happening there.

It is too big bite for anyone to migrate all tests at once. Therefore I tried to kick off a new structure using a modern tools, which seems simpler to use for me. I can imagine us incrementally migrating old tests to `tests_new` over time.


tests_new/test_geocaching.py
```diff
@@ -0,0 +1,178 @@
+import contextlib
```
tomasbedrich, line 1: Please note this file. Actually the whole `tests_new` directory is probably the most important change.

pyproject.toml
```diff
@@ -0,0 +1,48 @@
+[build-system]
```
tomasbedrich, line 1: This is probably the second most interesting thing – migration from `setup.py` to `pyproject.toml`.

CONTRIBUTING.rst
```diff
@@ -13,59 +13,97 @@ First time
...
-       Scripts\activate  # Windows
+       source .venv/bin/activate  # Unix
+       .venv\Scripts\activate  # Windows
 
+3. Write some **code and tests**. After that, don't forget to update the **docs**. See coding style below.
 
-3. Write some **code and tests**. After that, don't forget to update the **docs**. See coding style
-   below.
-4. Run the linter and tests:
+4. Format the code using `Black <https://black.readthedocs.io/en/stable/>`_,
```
tomasbedrich, line 54: And this is the last important thing – the choice of tools.

pycaching/geocaching.py
```diff
@@ -228,7 +228,7 @@ def logout(self):
         logging.info("Logging out.")
         self._logged_in = False
         self._logged_username = None
-        self._session = requests.Session()
+        self._session.cookies.clear()
```
FriedrichFroebel, line 231: I assume that this should fix the failing logout test?

FriedrichFroebel (commented):
From what I have seen, it looks good to me.

With the `tests_new` approach I am not sure whether we should still use the basic `unittest.TestCase` setup and the assertions from there instead of bare `assert` and the *pytest* contexts. Is there any reason for your approach? According to the [docs](https://docs.pytest.org/en/latest/how-to/unittest.html), *pytest* should be fine with *unittest* in theory (subtest support can be added with *[pytest-subtests](https://pypi.org/project/pytest-subtests/)*).

pycaching/geocaching.py
tomasbedrich: Not only – it was some weird combination of improper logout handling (this), incomplete mocking (see `test/__init__.py` change), usage of `setUpClass` (thus keeping the polluted Geocaching object available between test cases), executing multiple subtests in a single test method (the same issue with polluted objects) and last but not least confusing cassette naming and separation.

tomasbedrich:
Inheritance + mocking + Betamax cassettes proved itself to be too wild combination to maintain. Therefore I would rather get rid of classes at all, but if the only thing which remains is `unittest.TestCase` base class, I'm fine with that.

Re `assert`: it's a subjective preference. I like it because it's shorter than `self.assertXxx`, I don't need to remember special `Xxx` and `Yyy` for every type of assertion. And it plays well with pytest style tests (=simple top level functions).

Re subtest: Thanks, I haven't knew about that plugin! Anyway, I would rather get rid of subtests (and eventually turn them into [pytest parametrized tests](https://docs.pytest.org/en/6.2.x/parametrize.html)). The motivation is for subtests not to influence each other's state.

To sum it up and maybe provide some underlaying motivation: I am using `pytest` in my daily job for multiple years without any issues. 😄 Therefore I tend to reorganise tests the way I know they will work.

Sorry that not all my arguments are objective.

FriedrichFroebel:
Thanks for the explanations. Yes, it probably is a matter of taste and style incorporated in the daily workflow. For me class-based tests make it easier to detect test groups and the *unittest* asserts are more verbose in my opinion. We use *pytest* at work as well, but with some custom collection logic to be able to run some read-only tests on production systems. I am not fixed to any approach - as long as it works, I am fine with it, I asked just for curiosity. ;)

Issue #168, opened 2021-11-15
Fix docs build


tomasbedrich:
Due to `setup.py` removal, it started failing: https://readthedocs.org/projects/pycaching/builds/15273018/

(cc @FriedrichFroebel)

FriedrichFroebel:
I am not familiar enough with `pyproject.toml` (so I am not sure if this is the best solution), but we probably have to fix the version retrieval in `docs/conf.py` (tested with `make html`):

```python
# from setup import info
import toml
configuration = toml.load('../pyproject.toml')
```

and

```python
configured_version = configuration['project']['version']
# The short X.Y version.
version = ".".join(configured_version.split(".")[:2])
# The full version, including alpha/beta/rc tags.
release = configured_version
```

I am not sure about the version handling though, as the current release `4.2.2` leads to `4.2` in the docs.

As a side note: A warning is reported about some faulty docstring:

```
/usr/local/lib/python3.8/dist-packages/pycaching/cache.py:docstring of pycaching.cache.Cache.original_location:4: WARNING: Field list ends without a blank line; unexpected unindent.
```

Another side note: Why are we using this hardcoded Python 3.5 path logic, introduced in https://github.com/tomasbedrich/pycaching/commit/f4937db4242606d2cbab23baed4337dcd22ad88f? This should not have any effect in most of the cases.

```python
# import site-packages
for package_path in Path("../lib/python3.5/site-packages").glob("*.egg"):
    sys.path.append(str(package_path.resolve()))
```

tomasbedrich:
Fixing the way of reading project version looks like it may work. I am only afraid of using `toml`, as a dependency, which needs to be installed additionally.

Sorry, I'm not able to explain f4937db4242606d2cbab23baed4337dcd22ad88f.

What would I suggest is a bigger cut into the docs build. Maybe switching to Mkdocs? There is a [ready-made GitHub action](https://squidfunk.github.io/mkdocs-material/publishing-your-site/#with-github-actions) for publishing to GitHub pages. Does that sound interesting to you?

FriedrichFroebel:
[...]

PR #176, opened 2021-12-07
Fix docs build

FriedrichFroebel:
This fixes the docs build as discussed in #168. Warnings raised by Sphinx have been fixed as well, leading to the changes in the JSON snippets and the docstring.

This does not yet include any additional configuration like the additional RTD configuration file.

As already mentioned in #168, I am not sure whether the version modifications are correct, as the version display in the top left corner of the generated HTML page will be `4.2` at the moment, while the full version is `4.2.2`.


tomasbedrich (approved):

tomasbedrich:
It seems to be displayed correctly on https://pycaching.readthedocs.io/en/4.2.1/, so I would leave it as is. Thanks.

Issue #179, opened 2022-02-03
cache.load fails


tzutzanu:
I presume some changes on the geocaching site are causing this

```
>>> geocaching = pycaching.login()
>>> gc = 'GC9N1YP'
>>> cache = geocaching.get_cache(gc)
>>> cache.load_quick()
>>> cache.name
'Desert Island Discs: Easton Corbin'
>>> cache.load()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/pycaching/cache.py", line 725, in load
    self.name = cache_details.find("h2").text
AttributeError: 'NoneType' object has no attribute 'text'
```
cache.load_quick() seems to be ok.

FriedrichFroebel:
Thanks for the report.

From a quick look at it, only fixing the cache name retrieval will not be enough, as more places have been changed, with `_get_log_counts_from_cache_details` raising an exception as well.

Feel free to dig through the various places and submit a PR to get this library back to work.

PR #181, opened 2022-02-17
Issue 179

GTBoon72:
#179 This PR should fix this issue. Checked other self.name references, but they work fine as is.


FriedrichFroebel:
Thanks for the PR. Unfortunately your changes fail in the CI which seems to be an issue with the Betamax cassettes being outdated and requiring an update as well.

FriedrichFroebel:
The problem is not in the testing itself, but in the Geocaching.com site changes. For this reason your tests fail at the moment and you have to re-record the corresponding cassettes. Otherwise I will not be able to merge your changes.

GTBoon72:
Thanks for pointing out. Will see if I understand how that works, and will give it a try this weekend.

FriedrichFroebel:
The PM-only property can only be detected if you are logged in as a basic user, which is implicitly documented in the `load_by_guid` method, although this might not be obvious enough.

For basic members the print page of a PM-only cache will be mostly empty and contain the following snippet:

```html
        <div id="pnlErrorMessage">
	<p class="Warning">You must be a Premium Member to view this page.</p>
</div>
```

For non PM-only caches it will look like this:

```html
        <div id="pnlErrorMessage">

</div>
```

I will have a look at updating the test cassettes later. If you are okay with it and you have allowed reviewers to make changes to your pull request (and I manage to get this to work on my side), I can update the corresponding cassettes on your branch for you to check if your changes are okay or need a fix.

GTBoon72:
Hmm, yeah, that makes sense. But then, am I supposed to switch between PM and non-PM accounts, while running the tests? Did I overlook that in the test description?
Of course I am okay with your updates, it's your project after all! I enabled reviewers to update the code.
Thanks for your help!

tomasbedrich:
I am leaving that to you @FriedrichFroebel. Please ping me as soon as you need a release. Thanks!

CONTRIBUTING.rst
```diff
@@ -82,17 +82,18 @@ If you haven't written or modified any tests, tests can be run just like:
...
 worry, these will not leave your computer. Betamax will insert a placeholder when it records any
 new cassettes. To run new tests, first set up the following environment variables:
 
 .. code-block:: bash
 
-    PYCACHING_TEST_USERNAME="yourusername" PYCACHING_TEST_PASSWORD="yourpassword" pytest
+    PYCACHING_TEST_USERNAME="yourusername" PYCACHING_TEST_PASSWORD="yourpassword" pytest <test folder name>
 
 Substitute your username for ``yourusername`` and your password for ``yourpassword``.
+NB: for successful test results, you'll need to use a Basic Member account!
```
FriedrichFroebel, line 96: I would rather use something like

> This requires you to use a basic member account, otherwise you might see unexpected test failures.

CONTRIBUTING.rst
```diff
@@ -82,17 +82,18 @@ If you haven't written or modified any tests, tests can be run just like:
 
 .. code-block:: bash
 
-    pytest
+    pytest <test folder name>
```
FriedrichFroebel, line 85: Why did you change this? If you are in the root directory of the repository, it should work out of the box without specifying the directory.

GTBoon72: At first, that didn't work for me, it didn't execute any tests at all. Now that you mention it, I tried it without a default, and now it does work. Nonetheless, I think it is a useful addition, since it explains how to switch between the tests in the 'test' and 'tests_new' folders.

FriedrichFroebel (approved):
Thank you for your patience.

Issue #185, opened 2022-06-18
get_trackable does not work if no type is available


nagisml:
```
trackable = geocaching.get_trackable("TB7WZD9")
trackable = geocaching.get_trackable("TB7WVN8")
```

Others are working fine.

Here the traceback

```
Traceback (most recent call last):
  File "C:\Users\xyz\AppData\Local\Programs\Python\Python310\lib\site-packages\pycaching\util.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "C:\Users\xyz\AppData\Local\Programs\Python\Python310\lib\site-packages\pycaching\trackable.py", line 81, in name      
    return self._name
AttributeError: 'Trackable' object has no attribute '_name'. Did you mean: 'name'?
```

My quick solution was to add

```
if type is not None:
```

in the name and type defs.

nagisml:
fix the handling of none existign TB data in my fork and also added some TB feature.
https://github.com/nagisml/pycaching/tree/master/pycaching
Sure in not the best coding quality but it seems to work very well for me

FriedrichFroebel:
Thanks for the report. This appears to appear for generic trackable items only.

I am not sure about your

> in the name and type defs

comment though, as just checking for `if type is not None` before calling `type.strip()` should be enough in theory.

Feel free to add a pull request fixing this issue. For further features, please open a dedicated issue beforehand and a corresponding pull request, as this makes discussing and reviewing much easier.

I would be able to send a pull request to fix this issue as well, but I am currently trying to debug why my corresponding tests fail - the login reports to be successful, but when requesting the trackable page there is no logged-in user. Given that regular loading is successful, this seems to be related to Betamax somehow.

richardsondev:
Hey there, I ran across this exception as well and found this issue reported. I have raised a PR which fixed the exception for the TB I was looking at. Apparently this can happen for many older geocoins that aren't "Generic" as well.

PR #187, opened 2022-08-20
Handle missing TB type text like TB19G8B

richardsondev:
Reported in #185 

Ran into this with trackable TB19G8B which is missing the alt text on the icon image. 

This is adding a check to gracefully handle this case instead of throwing an exception.

### DIFF
--- a/pycaching/trackable.py
+++ b/pycaching/trackable.py
@@ -153,7 +153,10 @@ class Trackable(object):
 
     @type.setter
     def type(self, type):
-        self._type = type.strip()
+        if type is None:
+            self._type = ""
+        else:
+          self._type = type.strip()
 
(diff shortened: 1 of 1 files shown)


FriedrichFroebel:
Thank you for your PR, which seems to do the correct thing. There are two things though which prevent this PR from being merged:

1. You did not add any test for the correct behavior, at least for handling the error case, but it might be worth for the regular case as well.
2. *black* fails as it is too old, being incompatible with the latest *click* version. Although this is unrelated to your PR, would you mind bumping *black* to 22.6.0 (which is the latest release) to fix CI runs?

richardsondev:
Thanks for the guidance but I am not interested in further development on this. Thanks for understanding.

PR #192, opened 2022-08-23
more fields and logs from trackable page

nagisml:
Additional fields for TBs like origin, releaseDate
Provide the already loaded lastTBLogs from the first loaded page (up to ten logs).

### DIFF
--- a/pycaching/trackable.py
+++ b/pycaching/trackable.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python3
 
+from bs4 import BeautifulSoup
+
 from pycaching import errors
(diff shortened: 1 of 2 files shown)


FriedrichFroebel:
Thanks for the PR.

Doing some quick skimming through the code, there are some things which probably should be addressed:

* As documented in the contribution guide, please add tests (and docs if relevant) as well.
* Why are you re-parsing the source tree for retrieving the logs?
* We might want to move the log parsing functionality to a dedicated method to keep the parser methods more readable (see the corresponding methods for *Cache* objects for an example).

Issue #182, opened 2022-03-06
`Remote end closed connection without response` when using `Geocaching.search()`


BelKed:
When trying to load caches in some location using `Geocaching.search()`, it doesn't work because of error (`Remote end closed connection without response`).
However, loading a single geocache using `Geocaching.get_cache()` seems to work fine...

---

- **OS:** macOS Monterey `12.6.1`
- **Python version:** `3.9.2`
- **`pycaching` version:** `4.2.3`

---

**Python code with error:**
```py
import pycaching

geocaching = pycaching.login("Username", "Password")

point = geocaching.geocode("Prague")

for cache in geocaching.search(point, limit=10):
    print(cache.name)
```

---

**Error:**
```py
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 449, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 444, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/http/client.py", line 1374, in getresponse
    response.begin()
  File "/usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/http/client.py", line 318, in begin
    version, status, reason = self._read_status()
  File "/usr/local/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/http/client.py", line 287, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/usr/local/lib/python3.10/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    retries = retries.increment(
  File "/usr/local/lib/python3.10/site-packages/urllib3/util/retry.py", line 550, in increment
    raise six.reraise(type(error), error, _stacktrace)
[...]

PR #193, opened 2022-08-23
Add `Cache.status` property

### LINKED ISSUE #183: Cache.state does not correctly reflect the status of a geocache
The Cache.state variable does not correctly reflect the enabled status of a geocache. Geocaches that are disabled or archived return a state of True. The problem seems to be in the `load()` method:

```python3
        self.state = root.find("ul", "OldWarning") is None
```

Current geocache pages use `<div ...>` with `id="ctl00_ContentBody_disabledMessage"` or `id="ctl00_ContentBody_archivedMessage"`. Maybe the `load()` method should be:

```python3
self.state = not (root.find(id="ctl00_ContentBody_disabledMessage") or
                  root.find(id="ctl00_ContentBody_archivedMessage"))
```

This solution wouldn't correctly handle unpublished geocaches, but I don't know if that is a concern.

A geocache can have a status of Unpublished, Active, Disabled, Locked, or Archived. Would it be valuable to make all of these available in `Cache()`? Maybe through a new attribute, possibly `Cache.status`?

BelKed:
As discussed in #183, I'm opening a pull request to close this issue.

However, tests aren't written yet.
There's a test script for now:
```py
import pycaching

geocaching = pycaching.login("USERNAME", "PASSWORD")

caches = [
    "GC8CKQQ",  # Enabled
    "GC7Y77T",  # Disabled
    "GC1PAR2",  # Archived
    "GC10",     # Locked
]

for cache_wp in caches:
    cache = geocaching.get_cache(cache_wp)
    print(f"[https://coord.info/{cache.wp}] {cache.name} > {cache.status.name.capitalize()} • {cache.state}")
```

---

Fixes #183


BelKed:
Now I found that it doesn't work properly with the `Cache.load_quick()` method.
The received JSON states if the cache is available (it's not disabled nor locked) or if is archived. So, the locked status can't be determined. That means the complete load is needed. Or is it fine as it's now?

---

```
[https://coord.info/GC8CKQQ] Kohinoorka > Enabled • True
data['available'] = True
data['archived'] = False

[https://coord.info/GC7Y77T] Vojanovy sady > Disabled • False
data['available'] = False
data['archived'] = False

[https://coord.info/GC1PAR2] Kostel Panny Marie Ruzencove > Archived • False
data['available'] = False
data['archived'] = True

# Not actually archived, but locked
[https://coord.info/GC10] First Divine > Archived • False
data['available'] = False
data['archived'] = True
```

FriedrichFroebel:
Thanks for your PR.

It seems like the CI tests are currently failing, although I did not have a deeper look into it.

Did you have a look at how to parse the data from the print page (`_from_print_page`)? As far as I can see, the status will not be parsed there? Or is it not available from this page at all?

As far as I can see it, if the short-comings of the quick loading functionality are documented correctly, this should not be a real issue, unless @tomasbedrich thinks different about it.

pycaching/cache.py
```diff
@@ -853,7 +878,12 @@ def load_quick(self):
         # prettify data
         self.name = data["name"]
         self.type = Type.from_string(data["type"]["text"])
-        self.state = data["available"]
+
+        if data["available"]:
+            self.status = Status.enabled
+        else:
+            self.status = Status.archived if data["archived"] else Status.disabled
```
tomasbedrich, line 885: If we omit these two lines, full lazy load is triggered when user accesses `.status` -> filling the `.status` correctly. Right?

tomasbedrich:
Thanks for seeking for my input. My general preference would be _not_ to load attributes at all, if we are not able to load them correctly using given loading method. Then we can rely on lazy-loading.

pycaching/cache.py
BelKed: Good point 👍
Fixed in https://github.com/tomasbedrich/pycaching/pull/193/commits/43b3d3db9efae54cb47bd500709ad89999b5ed74

BelKed:
@FriedrichFroebel

> It seems like the CI tests are currently failing, although I did not have a deeper look into it.

They are failing because I marked the `Cache.state` property as read-only (I removed the `@setter` method for it) as it's dynamically set based on the `Cache.status`.
Some of the tests need to be updated.

> Did you have a look at how to parse the data from the print page (`_from_print_page`)? As far as I can see, the status will not be parsed there? Or is it not available from this page at all?

The current status it's not there. It even wasn't parsed earlier with `Cache.state`:
https://github.com/tomasbedrich/pycaching/blob/1f819261a7a89d2bf9799184e6db67cf76bc7c96/pycaching/cache.py#L148
I think it can be somehow guessed if we would also load logs (by adding `&lc=10`), but if there are more than 9 logs after the Archive log, it wouldn't be possible to find out.
So, I wouldn't burden with that...

> As far as I can see it, if the short-comings of the quick loading functionality are documented correctly, this should not be a real issue, unless @​tomasbedrich thinks different about it.

Already fixed based on his requirements in https://github.com/tomasbedrich/pycaching/pull/193/commits/43b3d3db9efae54cb47bd500709ad89999b5ed74

pycaching/cache.py
```diff
@@ -839,6 +863,7 @@ def load_quick(self):
         Use information from geocaching map tooltips. Therefore loading is very quick, but
         the only loaded properties are: `name`, `type`, `state`, `size`, `difficulty`, `terrain`,
         `hidden`, `author`, `favorites` and `pm_only`.
+        It also loads `status`, however locked caches are considered archived.
```
tomasbedrich, line 866: This needs to be updated after recent adjustment.

tomasbedrich (approved):

pycaching/cache.py
BelKed: Updated in https://github.com/tomasbedrich/pycaching/pull/193/commits/432268b4ad5fd9e14a6f2e1411134ed2fb9427cf

BelKed:
**Update:** I fixed the tests and added some new ones, so CI should run now without any problems.

BelKed:
Sorry, I forgot to sort the imports correctly...

test/test_cache.py
```diff
@@ -115,6 +115,12 @@ def test_waypoints(self):
     def test_state(self):
         self.assertEqual(self.c.state, True)
 
+    def test_status(self):
```
FriedrichFroebel, line 118: These tests seem to cover only the basic functionality. I would prefer if we would have a dedicated method which checks the correct parsing for all the available status values (except *unpublished* as this is hard to record/check).

pycaching/geocaching.py
```diff
@@ -289,7 +289,8 @@ def search(self, point, limit=float("inf")):
                 badge = row.find("svg", class_="badge")
                 c.found = "found" in str(badge) if badge is not None else False
                 c.favorites = row.find(attrs={"data-column": "FavoritePoint"}).text
-                c.state = not (row.get("class") and "disabled" in row.get("class"))
+                if not (row.get("class") and "disabled" in row.get("class")):
```
FriedrichFroebel, line 292: Did you add a test for this as well?

pycaching/cache.py
```diff
@@ -853,7 +878,12 @@ def load_quick(self):
         # prettify data
         self.name = data["name"]
         self.type = Type.from_string(data["type"]["text"])
-        self.state = data["available"]
+
+        # We can fill in status correctly only for enabled caches
```
FriedrichFroebel, line 882: Did you add a test for this as well?

pycaching/cache.py
```diff
@@ -780,7 +792,17 @@ def load(self):
 
         self.location = Point.from_string(root.find(id="uxLatLon").text)
 
-        self.state = root.find("ul", "OldWarning") is None
+        # Cache status
```
FriedrichFroebel, line 795: I would probably prefer this parsing to happen in a `classmethod` of `Status`, like

```
@classmethod from_cache_details(cls, root):
    ...
```

This would reduce the amount of code for the generic parsing method to make everything more readable.

BelKed: Done in https://github.com/tomasbedrich/pycaching/pull/193/commits/87dcf189a120198f050d8c7f2235e32fcc2ed3e3

test/test_cache.py
BelKed: Added in https://github.com/tomasbedrich/pycaching/pull/193/commits/b9a643fc43ba795e689cb32587d1dbda5c5096cd

pycaching/cache.py
BelKed: Now I'm trying and I ran into this issue:

<details>
    <summary>Stack trace</summary>

```
____________________________________________ TestCacheStatus.test_cache_status ____________________________________________

self = <pycaching.geocaching.Geocaching object at 0x10dd52eb0>, url = 'http://tiles01.geocaching.com/map.details'
expect = 'json', method = 'GET', login_check = True, kwargs = {'params': {'i': 'GC8CKQQ'}}, res = <Response [500]>

    def _request(self, url, *, expect="soup", method="GET", login_check=True, **kwargs):
        """
        Do a HTTP request and return a response based on expect param.
    
        :param str url: Request target.
        :param str method: HTTP method to use.
        :param str expect: Expected type of data (either :code:`soup`, :code:`json` or :code:`raw`).
        :param bool login_check: Whether to check if user is logged in or not.
        :param kwargs: Passed to `requests.request
            <http://docs.python-requests.org/en/latest/api/#requests.request>`_ as is.
        """
        # check login unless explicitly turned off
        if login_check and not self._logged_in:
            raise NotLoggedInException("Login is needed.")
    
        url = url if "//" in url else urljoin(self._baseurl, url)
    
        try:
            res = self._session.request(method, url, **kwargs)
>           res.raise_for_status()

pycaching/geocaching.py:83: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Response [500]>

    def raise_for_status(self):
        """Raises :class:`HTTPError`, if one occurred."""
    
        http_error_msg = ""
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode("utf-8")
            except UnicodeDecodeError:
                reason = self.reason.decode("iso-8859-1")
        else:
            reason = self.reason
    
        if 400 <= self.status_code < 500:
            http_error_msg = (
                f"{self.status_code} Client Error: {reason} for url: {self.url}"
            )
    
        elif 500 <= self.status_code < 600:
            http_error_msg = (
                f"{self.status_code} Server Error: {reason} for url: {self.url}"
            )
    
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: https://tiles01.geocaching.com/map.details?i=GC8CKQQ

.venv/lib/python3.8/site-packages/requests/models.py:1021: HTTPError

The above exception was the direct cause of the following exception:

self = <test.test_cache.TestCacheStatus testMethod=test_cache_status>

    def test_cache_status(self):
        with self.subTest("Enabled"):
            cache = Cache(self.gc, "GC8CKQQ")
            with self.recorder.use_cassette("cache_status_enabled"):
                self.assertEqual(Status.enabled, cache.status)
    
        with self.subTest("Disabled"):
            cache = Cache(self.gc, "GC7Y77T")
            with self.recorder.use_cassette("cache_status_disabled"):
                self.assertEqual(Status.disabled, cache.status)
    
        with self.subTest("Archived"):
            cache = Cache(self.gc, "GC1PAR2")
            with self.recorder.use_cassette("cache_status_archived"):
                self.assertEqual(Status.archived, cache.status)
    
        with self.subTest("Locked"):
            cache = Cache(self.gc, "GC10")
            with self.recorder.use_cassette("cache_status_locked"):
                self.assertEqual(Status.locked, cache.status)
    
        with self.subTest("Enabled > Quick load"):
            cache = self.gc.get_cache("GC8CKQQ")
>           cache.load_quick()

test/test_cache.py:498: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pycaching/cache.py:860: in load_quick
    res = self.geocaching._request(self._urls["tiles_server"], params={"i": self.wp}, expect="json")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pycaching.geocaching.Geocaching object at 0x10dd52eb0>, url = 'http://tiles01.geocaching.com/map.details'
expect = 'json', method = 'GET', login_check = True, kwargs = {'params': {'i': 'GC8CKQQ'}}, res = <Response [500]>

    def _request(self, url, *, expect="soup", method="GET", login_check=True, **kwargs):
        """
        Do a HTTP request and return a response based on expect param.
    
        :param str url: Request target.
        :param str method: HTTP method to use.
        :param str expect: Expected type of data (either :code:`soup`, :code:`json` or :code:`raw`).
        :param bool login_check: Whether to check if user is logged in or not.
        :param kwargs: Passed to `requests.request
            <http://docs.python-requests.org/en/latest/api/#requests.request>`_ as is.
        """
        # check login unless explicitly turned off
        if login_check and not self._logged_in:
            raise NotLoggedInException("Login is needed.")
    
        url = url if "//" in url else urljoin(self._baseurl, url)
    
        try:
            res = self._session.request(method, url, **kwargs)
            res.raise_for_status()
    
            # return bs4.BeautifulSoup, JSON dict or raw requests.Response
            if expect == "soup":
                return bs4.BeautifulSoup(res.text, "html.parser")
            elif expect == "json":
                return res.json()
            elif expect == "raw":
                return res
    
        except requests.exceptions.RequestException as e:
            if e.response.status_code == 429:  # Handle rate limiting errors
                raise TooManyRequestsError(
                    url, rate_limit_reset=int(e.response.headers.get("x-rate-limit-reset", "0"))
                ) from e
    
>           raise Error("Cannot load page: {}".format(url)) from e
E           pycaching.errors.Error: Cannot load page: http://tiles01.geocaching.com/map.details

pycaching/geocaching.py:99: Error
```

</details>

The test code is following:
```py
        with self.subTest("Enabled > Quick load"):
            cache = Cache(self.gc, "GC8CKQQ")
            cache.load_quick()
            with self.recorder.use_cassette("cache_status_enabled_load_quick"):
                self.assertEqual(Status.enabled, cache.status)
```

---

But this works without any problems:

```py
import pycaching
from pycaching import Cache

geocaching = pycaching.login("USERNAME", "PASSWORD")

cache = Cache(geocaching, "GC8CKQQ")
cache.load_quick()
print(cache.status)
```

I can also load the [URL](https://tiles01.geocaching.com/map.details?i=GC8CKQQ) in the browser and it shows the needed JSON.

pycaching/geocaching.py
BelKed: Now I'm trying to add tests for this as well, and I forgot there's some problem with the `Geocaching.search()` method:
- https://github.com/tomasbedrich/pycaching/issues/182

BelKed: Now I'm trying to add tests for this as well, and I forgot there's some problem with the `Geocaching.search()` method:
- https://github.com/tomasbedrich/pycaching/issues/182

That means I can't record cassettes...

pycaching/cache.py
FriedrichFroebel: There seems to be some issue with the cookies sent by the application:

```
[HttpRequestValidationException]: A potentially dangerous Request.Cookies value was detected from the client (gspkauth=&quot;&lt;AUTH COOKIE&gt;&quot;)
```

Nevertheless, I am not sure why this is the case here. I am going to open a corresponding issue.

For the time being, just uncomment the remaining code of the corresponding test class and record the cassette, then remove the comment markers again. Additionally, your code should call `cache.load_quick()` inside the recorder context instead of doing it beforehand, as the cassette would be empty otherwise.

pycaching/geocaching.py
FriedrichFroebel: Sorry, I did not remember this. In this case, we probably have to omit the corresponding test case here.

pycaching/cache.py
BelKed: Thanks for the help :)
I added it in https://github.com/tomasbedrich/pycaching/pull/193/commits/cd51a3e84245c4951ab943de2c486d7250f6c51c

FriedrichFroebel (approved):
Thanks for the heads up.

PR #200, opened 2022-12-29
add second unarchive log id

serycjon:
encountered here: https://www.geocaching.com/geocache/GCGN30


FriedrichFroebel:
Thanks for the PR. Could you please add a corresponding test as well?

serycjon:
Not sure if this is what you meant ^.

FriedrichFroebel:
Thanks. For me this appears to be sufficient for now.

FriedrichFroebel (approved):

PR #201, opened 2023-01-09
Add more date patterns

BelKed:
It seems that the date format no longer depends on the language settings. In [`Settings > Preferences`](https://www.geocaching.com/account/settings/preferences), you can find a `Date Format` section, where you can set a variety of date formats:
<p align=center>
Screenshot
</p>

---

This PR adds support for all of them.
Related to stale #155.


FriedrichFroebel:
Thanks for the updated PR. Is there any reason for dropping the `г` format completely, whereas only Windows is affected (maybe related to https://bugs.python.org/issue8304)? We already have exceptions for Windows date formats inside our code and should probably use a similar approach here: https://github.com/tomasbedrich/pycaching/blob/2a6067bb6340eb23a41fd6f8d7789645fc155e4f/pycaching/util.py#L99

Just for the record, the offending exceptions is the following one:

```python3
>           formatted_date = datetime.datetime.strftime(date, pattern)
E           UnicodeEncodeError: 'locale' codec can't encode character '\u0433' in position 9: encoding error
```

BelKed:
Just the Windows build was failing…
I’ll take a look at it in the evening and hopefully add a commit with fix for Windows :)

BelKed:
As I promised, I fixed it...

Feel free to let me know about any issues with this solution :)

FriedrichFroebel:
Thanks, seems legit to me. Nevertheless, I would probably prefer to avoid this overhead for Linux and macOS. Could you make the offending section conditional, please? Meaning:

```python3
if platform.system() == "Windows":
    # Use solution with encoding and decoding.
else:
   # Use old solution.
```

BelKed:
Sure, https://github.com/tomasbedrich/pycaching/pull/201/commits/8e2acd2d2fcd8e0501ee3fa91e673db9e0de5ab1

FriedrichFroebel (approved):
Thanks for your fix and patience.

Issue #180, opened 2022-02-11
Use new v2 of search api


CachingFoX:
Groundspeak provides a new web site and API for search. The old search API is currently still available in the background.
[https://www.geocaching.com/api/proxy/web/search/v2?skip=0&take=50&asc=False&sort=favoritepoint](https://www.geocaching.com/api/proxy/web/search/v2?skip=0&take=50&asc=False&sort=favoritepoint)
The new API returns JSON data instead of pre-render HTML code with more info (see picture) 

The idea implement a new function to use the new version of v2.

![grafik](https://user-images.githubusercontent.com/1838534/153613426-6b46e915-87b0-4167-ad38-3c5da5f2b4a0.png)

CachingFoX:
I will implement a new search method. Currently not all features will be supported well by www.geocaching.com.

Search arguments will not be parsed or verified. Arguments are send directly to GC. 

Sorting options: TBD. 

Feature have to be tested for Basic and Premium members.

The method returns (in default) a generator for Cache objects. The generator doesn't trigger the load() method of Cache. The generator fills the created Cache object with all available data returned by the search.

For advanced usage the structure of the search methid can be configured (raw text, parsed list/dict, every cache as dict(generator)).

Limitation:no limitations are planned

PR #175, opened 2021-12-05
Implementation Advanced Search

CachingFoX:
The PR provides an interface to the advanced search of www.geocaching.com. The new method can be use the full power of advanced search.

**Features**
* Powerful search with a lot of filters
* Results are sortable (asc and desc)
* Limit number of returned caches
* Solve issue #148 (caches owned by an users) 
* Old search method still exists and use internally the new one
* Full usage of AJAX endpoint

**Known Limitations, Problems Bugs**
* Works only with the English website (for multi language support I18N helper from #170 required)
* There is not hint in the AJAX response if a search options is only available for PM users. GS perform a normal search :-/

Issue #148, opened 2020-05-11
Searching for user hides


pkubiak:
Hi,
It is possible using `pycaching` to search for caches hidden by given user (e.g. https://www.geocaching.com/seek/nearest.aspx?u=Geocaching+HQ) ? If not I will be interested in implementing such a feature :)

FriedrichFroebel:
No, this is not implemented for now. There already is a similar issue, see #143.

Issue #143, opened 2020-04-22
Advanced the search method


CachingFoX:
I want to advanced the search function. Currently the search method is very limited. The search on the web page offers a lot of filter parameters, which are not supported by the search method.

I have already implement a prototype of advanced search. But the quality is not good enough for a pull request. The prototype is not fully featured, no test cases at the moment and some open questions.

Parts of implementation of the search method should be reused and enhanced. But current search method should still exists for backward compatibility, and share the implementation with the advanced search.

Filter parameters
* Worldwide or origin with radius
* Geocache types (and group of types)
* Found status
* Owner ship 
* Hidden by
* Terrain and Difficulty ranking in ranges
* Geocache size
* Membership type
* Cache status
* Corrected Coordinates
* Personal Note
* Place Date (after, before, between, on)
* Keyword
* Not found by
* Limit to country or region (includes a enumeration of countries and regions)
* Favorite points

tomasbedrich:
Unfortunately, most of filters you mention are available to PM only. I'm in for implementing these, but I don't have PM so I cannot help with this.

So if you want, keep working on this but keep in mind, that probably nobody will be able to help you (not even by review). :( Also please note, that we must keep the pycaching project fully functional for basic members.

CachingFoX:
> Unfortunately, most of filters you mention are available to PM only. I'm in for implementing these, but I don't have PM so I cannot help with this.
Sorry, I'm PM for a so long time, I forgot from time to time BM has a limit access to the web page. 
My solution has less modifications of the existing search function. This modification includes mainly extended handling of parameters in the URL, no changes in behaviour of loading results and more results.
Build the URL parameters for a filter is done in an extra class, because the class has a lot of simple getter/setter function and helper function to build parameters. If I implement this in the geocaching class or in the search function, it will kill the code.

During the days I improve my implementation as much as possible and create a lot of test which verifiy the build of the parameters. 

With the search function we have a similar problem as with the Personal Note. If someone re-record the cassette he will get a different result: new and missing caches, different sort order, different values (like favorite points), ...

In the next days I will send you a link to my implementation and you can give me feedback, about the architecture of my implementation. Interface good or bad.

I write a little how to use the new search function.

> So if you want, keep working on this but keep in mind, that probably nobody will be able to help you (not even by review). :( Also please note, that we must keep the pycaching project fully functional for basic members.

I will check what happened if a BM try to do PM features in the search

PR #206, opened 2023-03-25
Implementation of advanced search

BelKed:
This PR implements a new way of searching for caches using the search API, which allows using filters and much more (with the `Geocaching.advanced_search()` method).

This search method is also faster because it loads less data (the data is in JSON form). Data parsing is then easier and clearer.

---

Now we have a main search function — `Geocaching.advanced_search()`.
Both `Geocaching.search()` and `Geocaching.search_rect()` internally use `Geocaching.advanced_search()`, which reduces repeating code, since the implementation would be almost the same.

I've also updated the docstrings for those functions and added some tests.

---

Related PR:
- Closes https://github.com/tomasbedrich/pycaching/pull/175
(Although, that PR uses a different approach by parsing HTML...)

---

Related issues:
- Fixes https://github.com/tomasbedrich/pycaching/issues/143
- Fixes https://github.com/tomasbedrich/pycaching/issues/148
- Fixes https://github.com/tomasbedrich/pycaching/issues/180


test/test_geocaching.py
tomasbedrich, line 1: What do you think about taking this opportunity and migrating all the search-related tests to `tests_new`?

pycaching/geocaching.py
```diff
@@ -383,55 +316,111 @@ def search_quick(self, area, *, strict=False, zoom=None):
         #             # can yield more caches (which are not exactly in desired area)
         #             yield cache
 
-    # add some shortcuts ------------------------------------------------------
-
     def search_rect(
         self,
         rect: Rectangle,
-        *,
```
tomasbedrich, line 391: My preference is to keep the asterisk there. This way we ensure people pass the "less important" params always kw-based, which gives us flexibility to shuffle them around & possibly add a new one in the middle.

pycaching/geocaching.py
```diff
@@ -239,116 +239,49 @@ def get_logged_user(self, login_page=None):
...
-    def _search_get_page(self, point, start_index):
-        """Return one page for standard search as class:`bs4.BeautifulSoup` object.
-
-        :param .geo.Point point: Search center point.
-        :param int start_index: Determines the page. If start_index is greater than zero, this
-            method will use AJAX andpoint which is much faster.
+    def search(
+        self,
+        point: Point,
+        per_query: int = 200,
```
tomasbedrich, line 245: Re-ordering previous `point, limit` args breaks API for those using positional arguments only. **This is a blocking note.**

pycaching/geocaching.py
```diff
@@ -239,116 +239,49 @@ def get_logged_user(self, login_page=None):
...
+        sort_by: Union[str, SortOrder] = SortOrder.date_last_visited,
```
tomasbedrich, line 246: My preference is against adding new parameters to this method – to keep it dead simple for basic users and onboarding purposes.

On the other hand, I see some benefit of unifying the param list with `search_rect`. In that case, I'd strongly prefer having `*` here (see my other comment).

Given I am not daily user of this library anymore, so I will let others decide here.

tomasbedrich (changes requested):

tomasbedrich:
Nice job overall. 💪🏼 It would definitely be useful to have another pair of eyes here.

test/test_geocaching.py
BelKed: I guess I would keep that for another PR...

pycaching/geocaching.py
BelKed: Okay, good reason, resolved in https://github.com/tomasbedrich/pycaching/pull/206/commits/f3c20cb99b05f020d91a241936d3c68b9d674fc0

pycaching/geocaching.py
BelKed: Oh, didn't think about that, fixed it in https://github.com/tomasbedrich/pycaching/pull/206/commits/340afb8fbfbdfc4232c9536a00b9da45d84ccccf

pycaching/geocaching.py
BelKed: I still didn't quite get how the asterisk works there 😄

However, as an active library user, I would say I like having those arguments there since they make the work with caches a little bit easier. Those other arguments are optional, so users don't need to specify them.

The function needs to get only specified the `point` parameter. (`geocaching.search(Point(49.73, 13.38))` would work without any problems)

So, I hope adding the mentioned asterisk would be fine — https://github.com/tomasbedrich/pycaching/pull/206/commits/ea4817471fe5dc1b930be429f4aeeff8e3bf4560

test/test_geocaching.py
```diff
@@ -28,15 +28,15 @@ def test_my_dnfs(self):
 
     def test_search(self):
         with self.subTest("normal"):
-            tolerance = 2
-            expected = {"GC5VJ0P", "GC41FJC", "GC50AQ6", "GC167Y7", "GC7RR74", "GC167Y7"}
+            tolerance = 3
```
FriedrichFroebel, line 31: Do we have some way to avoid this tolerance value to further grow?

FriedrichFroebel (changes requested):
Thanks for the PR. I have marked some additional oddities which probably should be resolved before merging.

pycaching/geocaching.py
```diff
@@ -239,116 +239,50 @@ def get_logged_user(self, login_page=None):
...
+        limit: int = float("inf"),
+        sort_by: Union[str, SortOrder] = SortOrder.date_last_visited,
+        reverse: bool = False,
+        per_query: int = 200,
+        wait_sleep: bool = True,
+    ) -> Generator[Optional[Cache], None, None]:
+        """Search for caches around a specified location using a search API.
+
+        :param point: The :class:`.geo.Point` object representing the center point of the search.
+        :type point: `geo.Point`
```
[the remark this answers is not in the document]
BelKed: Resolved in https://github.com/tomasbedrich/pycaching/pull/206/commits/68e5a2ee8a0c94432e3ea8e8cca6c6b91a054a10

pycaching/geocaching.py
```diff
@@ -383,55 +317,112 @@ def search_quick(self, area, *, strict=False, zoom=None):
...
+        """Perform an advanced search for geocaches with specific search criteria.
+
+        The search is performed using the options provided in the :code:`options` parameter.
+        Example of the :code:`options` parameter::
+
+            # https://www.geocaching.com/play/search?owner[0]=Geocaching%20HQ&a=0
+            options = {"owner[0]": "Geocaching HQ", "a": "0"}
+
+        :param options: A dictionary of search options.
+        :type options: dict
```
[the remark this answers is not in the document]
BelKed: Resolved in https://github.com/tomasbedrich/pycaching/pull/206/commits/68e5a2ee8a0c94432e3ea8e8cca6c6b91a054a10

pycaching/geocaching.py
```diff
@@ -383,55 +317,112 @@ def search_quick(self, area, *, strict=False, zoom=None):
...
-        :param limit: Maximum number of caches to return.
-        :param origin: Origin point for search by distance.
-        :param wait_sleep: In case of rate limits exceeding, wait appropriate time if set True,
-            otherwise just yield None.
+        wait_sleep: bool = True,
+    ) -> Generator[Optional[Cache], None, None]:
+        """Search for caches in a specified :class:`.Rectangle` area using a search API.
+
+        :param rect: The :class:`.Rectangle` object representing the search area.
+        :type rect: geo.Rectangle
```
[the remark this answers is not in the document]
BelKed: Resolved in https://github.com/tomasbedrich/pycaching/pull/206/commits/68e5a2ee8a0c94432e3ea8e8cca6c6b91a054a10

test/test_geocaching.py
BelKed: Yup, resolved in https://github.com/tomasbedrich/pycaching/pull/206/commits/8534729f50fae8825939683d2410606ac22ecae2

BelKed:
Weirdly, CI fails here, but it runs on my fork:
https://github.com/BelKed/pycaching/actions/runs/4533795088

FriedrichFroebel:
I am not completely sure about the CI failures as well, but I do not really feel like this should be merged until we are able to resolve this. Maybe rebasing the code on the latest master change something here?

BelKed:
I believe it's fine now :)

FriedrichFroebel:
Thanks, looks good for me. Re-requesting review from @tomasbedrich before merging.

pycaching/geocaching.py
BelKed: Oh, now I get how it works :D
So, yeah, it makes sense now :)

Fixed in: https://github.com/tomasbedrich/pycaching/pull/206/commits/8c839f455920536d376fc9e8665663a19f3aa6f2. I've also reverted my changes in the test back to test whether positional arguments work: https://github.com/tomasbedrich/pycaching/pull/206/commits/cc0c92f1ef9504f079ed9cfe988a7e16a85caac7.

tomasbedrich (approved):

Issue #216, opened 2023-09-04
No current way to get the raw HTML cache description


DarkOutcast6:
There is no current way to get a raw HTML cache description (at least that I know)

FriedrichFroebel:
This is correct. Feel free to submit a PR for adding a new `description_html` attribute.

DarkOutcast6:
What library do you use to strip the geocaching website? I’ll try my best to find a way to get the raw HTML version of the description.

DarkOutcast6:
Just made a pull request for this issue. Pull request number is #218.
It would be great if you could explain what I need to add to make it be usable for the user.

Thank you so much for the help, though!

FriedrichFroebel:
By adding a simple test you should be able to ensure that the attribute can indeed be accessed and contains the correct data. Documentation is already present, so at least in theory this should be sufficient to make it usable.

PR #218, opened 2023-09-06
Adding html_description attribute to the Cache object

DarkOutcast6:
As stated in issue #216, you cannot currently acces the raw HTML version of a description.
So, I'm working on adding a `html_description` attribute to the Cache object. I understand it may not work, but I'm semi-confident I understand Beautiful Soup well enough and added the right code. Bear with me though, as the code could definitely contain mistakes.

### DIFF
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -829,7 +829,7 @@ class Cache(object):
 
         self.summary = root.find(id="ctl00_ContentBody_ShortDescription").text
         self.description = root.find(id="ctl00_ContentBody_LongDescription").text
-        self.descriptin_html = root.find(id="ctl00_ContentBody_LongDescription")
+        self.description_html = root.find(id="ctl00_ContentBody_LongDescription")
         
         self.hint = rot13(root.find(id="div_hint").get_text(separator="\n"))
 
@@ -955,7 +955,7 @@ class Cache(object):
 
         self.description = content.find("h2", string="Long Description").find_next("div").text
 
-        self.descriptin_html = content.find("h2", string="Long Description").find_next("div")
+        self.description_html = content.find("h2", string="Long Description").find_next("div")
 
         self.hint = content.find(id="uxEncryptedHint").get_text(separator="\n")
 
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -223,6 +223,7 @@ class Cache(object):
             "attributes",
             "summary",
             "description",
+            "html_description",
             "hint",
             "favorites",
             "pm_only",
@@ -616,6 +617,20 @@ class Cache(object):
 
     @property
     @lazy_loaded
+    def html_description(self):
+        """The cache long description in raw HTML.
+
+        :type: :class:`str`
+        """
+        return self._description
+
+    @description.setter
+    def html_description(self, description):
+        description = str(description).strip()
+        self._html_description = description
+
+    @property
+    @lazy_loaded
     def hint(self):
         """The cache hint.
 
@@ -813,7 +828,8 @@ class Cache(object):
 
         self.summary = root.find(id="ctl00_ContentBody_ShortDescription").text
         self.description = root.find(id="ctl00_ContentBody_LongDescription").text
-
+        self.html_description = root.find(id="ctl00_ContentBody_LongDescription")
+        
         self.hint = rot13(root.find(id="div_hint").get_text(separator="\n"))
 
         favorites = root.find("span", "favorite-value")
@@ -937,6 +953,8 @@ class Cache(object):
         self.summary = content.find("h2", string="Short Description").find_next("div").text
 
         self.description = content.find("h2", string="Long Description").find_next("div").text
+
+        self.html_description = content.find("h2", string="Long Description").find_next("div")
 
         self.hint = content.find(id="uxEncryptedHint").get_text(separator="\n")
 
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -1022,7 +1022,7 @@ class Cache(object):
         if not element:
             raise errors.ValueError("Log counts could not be found.")
 
-        # Text gives numbers and verbose s of the current values as well as an
+        # Text gives numbers and verbose of the current values as well as an
         # introductory text. So we have to perform number checks for each element and only keep
         # the numbers.
         # The values might contain thousand separators, which we have to remove before converting
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -1022,7 +1022,7 @@ class Cache(object):
         if not element:
             raise errors.ValueError("Log counts could not be found.")
 
-        # Text gives numbers and verbose of the current values as well as an
+        # Text gives numbers and verbose descriptions of the current values as well as an
         # introductory text. So we have to perform number checks for each element and only keep
         # the numbers.
         # The values might contain thousand separators, which we have to remove before converting
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -155,6 +155,7 @@ class Cache(object):
             del cache_info["attributes"]["attribute"]
         cache_info["summary"] = content.find("h2", string="Short Description").find_next("div").text
         cache_info["description"] = content.find("h2", string="Long Description").find_next("div").text
+        cache_info["html_description"] = content.find("h2", string="Long Description").find_next("div")
         hint = content.find(id="uxEncryptedHint")
         cache_info["hint"] = hint.get_text(separator="\n") if hint else None
         cache_info["waypoints"] = Waypoint.from_html(content, table_id="Waypoints")
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -830,7 +830,7 @@ class Cache(object):
         self.summary = root.find(id="ctl00_ContentBody_ShortDescription").text
         self.description = root.find(id="ctl00_ContentBody_LongDescription").text
         self.description_html = root.find(id="ctl00_ContentBody_LongDescription")
-        
+
         self.hint = rot13(root.find(id="div_hint").get_text(separator="\n"))
 
         favorites = root.find("span", "favorite-value")
@@ -1098,7 +1098,6 @@ class Cache(object):
                 return
 
             for log_data in logbook_page:
-
                 limit -= 1  # handle limit
                 if limit < 0:
                     return
@@ -1140,7 +1139,6 @@ class Cache(object):
         names = [re.split(r"[\<\>]", str(link))[2] for link in links if "track" in link.get("href")]
 
         for name, url in zip(names, urls):
-
             limit -= 1  # handle limit
             if limit < 0:
                 return
--- a/pycaching/cache.py
+++ b/pycaching/cache.py
@@ -154,8 +154,9 @@ class Cache(object):
         if "attribute" in cache_info["attributes"]:  # 'blank' attribute
             del cache_info["attributes"]["attribute"]
         cache_info["summary"] = content.find("h2", string="Short Description").find_next("div").text
-        cache_info["description"] = content.find("h2", string="Long Description").find_next("div").text
-        cache_info["description_html"] = str(content.find("h2", string="Long Description").find_next("div"))
+        raw-description = content.find("h2", string="Long Description").find_next("div")
+        cache_info["description"] = long-description.text
+        cache_info["description_html"] = str(long-description)
         hint = content.find(id="uxEncryptedHint")
         cache_info["hint"] = hint.get_text(separator="\n") if hint else None
         cache_info["waypoints"] = Waypoint.from_html(content, table_id="Waypoints")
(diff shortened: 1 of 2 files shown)


FriedrichFroebel:
Thanks for the test. Could you please add a check for the actual parsing from the HTML as well by extending one of the existing cassette-based tests? Additionally, CI is currently failing due to code style violations.

Do you have a solution for this?

> Additionally, do you see any way to avoid doing the same tree querying twice for both the `description` and `description_html` attribute?

DarkOutcast6:
I'm not entirely sure what you mean by the following. (I apologize, as I am a novice with these type of things)
> Could you please add a check for the actual parsing from the HTML as well by extending one of the existing cassette-based tests?

------------------------------

I think I understand what you mean by 
> avoid doing the same tree querying twice for both the description and description_html attribute?

If I _do_ understand you correctly, I'm working on it now.

DarkOutcast6:
I'm not entirely sure what you mean by the following. (I apologize, as I am a novice with these type of things)
> Could you please add a check for the actual parsing from the HTML as well by extending one of the existing cassette-based tests?

------------------------------

I think I undertsnad what you mean by 
> avoid doing the same tree querying twice for both the description and description_html attribute?

If I _do_ understand you correctly, I'm working on it now.

FriedrichFroebel:
> I'm not entirely sure what you mean by the following. (I apologize, as I am a novice with these type of things)
 >>   Could you please add a check for the actual parsing from the HTML as well by extending one of the existing cassette-based tests?

Your current test only covers explicitly setting the `description_html` field in one case, but does not ensure that the parsed data from the Geocaching.com page is indeed correct.  AFAIK this is not tested for the `description` field either, but checking this makes sense.

> I think I understand what you mean by
>> avoid doing the same tree querying twice for both the description and description_html attribute?
> If I do understand you correctly, I'm working on it now.

You only fixed this for exactly one case, but not for the remaining ones where the same approach might make sense as well.

Additionally, when you look at the CI runs here, you will see that there currently are code style violations. And https://github.com/tomasbedrich/pycaching/pull/218/files#diff-c1861c7962a2fda2a943282daf8c8cb0030863a5cfdf8acc43402e3f1b381857R157 is considered both a name violation according to PEP8 and invalid variable name syntax.

DarkOutcast6:
Sorry for the trouble -- I feel so close :)

Pretty sure I figured out what the code violation is. (correct me if I didn't fix it or if there is more)

___

> You only fixed this for exactly one case, but not for the remaining ones where the same approach might make sense as well.

I looked over the code and can't find another case where this occurs.

___

> Your current test only covers explicitly setting the description_html field in one case, but does not ensure that the parsed data from the Geocaching.com page is indeed correct. AFAIK this is not tested for the description field either, but checking this makes sense.

I understand what you mean, but how would I actually do this?

___

Sorry for sounding being an idiot through this whole procces lol. Hope I'm not too annoying! :wink:

FriedrichFroebel:
> Pretty sure I figured out what the code violation is. (correct me if I didn't fix it or if there is more)

Your variable names should use an underscore, not a dash (this is invalid Python syntax). Additionally, CI is reporting an empty line too much. This needs to be fixed.

> I looked over the code and can't find another case where this occurs.

You added changes to https://github.com/tomasbedrich/pycaching/pull/218/files#diff-c1861c7962a2fda2a943282daf8c8cb0030863a5cfdf8acc43402e3f1b381857R832 and https://github.com/tomasbedrich/pycaching/pull/218/files#diff-c1861c7962a2fda2a943282daf8c8cb0030863a5cfdf8acc43402e3f1b381857R958 as well, not only at https://github.com/tomasbedrich/pycaching/pull/218/files#diff-c1861c7962a2fda2a943282daf8c8cb0030863a5cfdf8acc43402e3f1b381857L157 You should use the same approach all over.

> I understand what you mean, but how would I actually do this?

There are multiple tests in `test.test_cache.TestMethods`. Consider extending the tests for the three loader methods to test for `description` (if currently missing) and `description_html`, ensuring that the HTML variant includes HTML tags, while the plain-text one does not. https://github.com/tomasbedrich/pycaching/blob/bef4ca07693d4d1e06b070f4c72a154ca3f487a4/test/test_cache.py#L282 is an example of an existing simple `description` test where the existing assertion probably should be expanded to cover a case where HTML tags have been removed (which is not the case here yet).

DarkOutcast6:
What is the difference between ` self.assertIn()`, ` self.assertEqual()` and ` self.assertGreater()` and which of these should be used for description_html?

FriedrichFroebel:
I recommend you to familiarize yourself with general `unittest`-based testing (which is the stdlib Python module for testing) to do so. You will find the official documentation for the different assertions at https://docs.python.org/3/library/unittest.html#assert-methods

DarkOutcast6:
What else do I need to add for this to be valid?
It be very helpful if you could give me a list of things I need to edit.

FriedrichFroebel:
* The code style needs to be fixed. See the CI runs for this or run `black` in your local development environment to check/fix this.
* As mentioned in the second section at https://github.com/tomasbedrich/pycaching/pull/218#issuecomment-1712584664, please consider re-using the retrieved tree for the text and string rendering as you have done for the other case.
* As mentioned in the third section at https://github.com/tomasbedrich/pycaching/pull/218#issuecomment-1712584664, please extend the tests for the three loader methods you changed with suitable tests (the test you added in your latest commit would probably fail for this specific geocache instance). To do so, please have a look at the original descriptions of the corresponding geocaches used for testing purposes.
  * Add a corresponding assertion for a part of the `description_html` property, ensuring that it does indeed include the correct HTML tags.
  * Improve/rewrite the current assertion for the `description` property to check for exactly the same section as in `description_html` with the only difference being that the expected content does indeed not contain the corresponding HTML tags.

DarkOutcast6:
>The code style needs to be fixed.

I fixed the `black` code styling issues.

___

>  re-using the retrieved tree for the text and string rendering as you have done for the other case.

Made it so that it now re-uses the tree.

___

The CI runs now state that there is a problem with `flake8`. The error reads: 

./pycaching/cache.py:628:5: F8[11](https://github.com/tomasbedrich/pycaching/actions/runs/6166084369/job/16740063682?pr=218#step:7:12) redefinition of unused 'description_html' from line 619
    @description.setter
    ^
1     F811 redefinition of unused 'description_html' from line 619

I don't fully understand what it means. Could you explain what it means?

pycaching/cache.py
```diff
@@ -1020,7 +1022,7 @@ def _get_log_counts_from_print_page(soup):
         if not element:
             raise errors.ValueError("Log counts could not be found.")
 
-        # Text gives numbers and verbose descriptions of the current values as well as an
+        # Text gives numbers and verbose s of the current values as well as an
```
FriedrichFroebel, line 1025: There is a typo.

pycaching/cache.py
```diff
@@ -613,6 +616,20 @@ def description(self, description):
...
+    @property
+    @lazy_loaded
+    def description_html(self):
+        """The cache long description in raw HTML.
+
+        :type: :class:`str`
+        """
+        return self._description
+
+    @description.setter
```
FriedrichFroebel, line 628: This should be the setter for the `description_html` property. At the moment, you have two setters for `description`.

pycaching/cache.py
DarkOutcast6: I don’t understand — I did not alter line 1020 - 1025.

pycaching/cache.py
DarkOutcast6: Just fixed!

pycaching/cache.py
FriedrichFroebel: You can see from the referenced diff that `descriptions` has been there before which still should be kept.

pycaching/cache.py
```diff
@@ -153,8 +153,10 @@ def _from_print_page(cls, geocaching, guid, soup):
         cache_info["attributes"] = {attr_name: attr_setting == "yes" for attr_name, _, attr_setting in attributes}
         if "attribute" in cache_info["attributes"]:  # 'blank' attribute
             del cache_info["attributes"]["attribute"]
-        cache_info["summary"] = content.find("h2", string="Short Description").find_next("div").text
-        cache_info["description"] = content.find("h2", string="Long Description").find_next("div").text
+        cache_info["summary"] = content.find("h2", string="Short ").find_next("div").text
```
FriedrichFroebel, line 156: Why did you change this string from "Short Description" to "Short "? This seems to be wrong and lets the existing tests fail.

FriedrichFroebel:
Thanks for the fixes. I have added two other comments to the diff. Additionally, the required tests are still missing or incomplete and should be implemented as requested previously.

pycaching/cache.py
DarkOutcast6: I didn’t. That’s so weird :/

pycaching/cache.py
DarkOutcast6: Fixed it!

pycaching/cache.py
```diff
@@ -613,6 +616,20 @@ def description(self, description):
...
         self._description = description
 
+    @property
+    @lazy_loaded
+    def description_html(self):
+        """The cache long description in raw HTML.
+
+        :type: :class:`str`
+        """
+        return self._description
```
FriedrichFroebel, line 626: I just have  been confused because of the CI results, but this should be `return self._description_html` of course.

DarkOutcast6: Just fixed! Thanks for pointing that typo out!

DarkOutcast6:
> Additionally, the required tests are still missing or incomplete and should be implemented as requested previously.

Working on this now, hopefully will be finished this week!
If there is anything else that needs to be brought to my attention, feel free to tell me now!

DarkOutcast6:
> Add a corresponding assertion for a part of the description_html property, ensuring that it does indeed include the correct HTML tags.
Just fixed it!

Hypothetically, this should work now, correct? If so, is there anything else I need to update? (And if it isn't, I'll update accordingly)

FriedrichFroebel:
You implemented the new property for the three different loading methods. I can find (at most) only two tests which check that loading the data does indeed work correctly for `description_html`. We should cover all three cases.

I previously requested that the corresponding tests for the `description` should be adopted to actually show that they do not contain HTML tags. This is not the case at the moment, thus we cannot see from the tests if `description` does indeed not contain any HTML tags, as there is nothing to strip for the existing text. Ideally, your test snippet for `description` and `description_html` would be identical with only the tags differing, while having HTML tags in-between as well, not only at the borders.

Additionally, linting currently fails due to the long line. Do you really need such a long text for the test or could you shorten this to still cover the relevant cases (see previous paragraph), but not being too verbose?

DarkOutcast6:
> description should be adopted to actually show that they do not contain HTML tags. This is not the case at the moment, thus we cannot see from the tests if description does indeed not contain any HTML tags, as there is nothing to strip for the existing text. Ideally, your test snippet for description and description_html would be identical with only the tags differing, while having HTML tags in-between as well, not only at the borders.
Will work on this now.

_____

> Additionally, linting currently fails due to the long line.
Fixing now.

____

> You implemented the new property for the three different loading methods. I can find (at most) only two tests which check that loading the data does indeed work correctly for description_html. We should cover all three cases.
Not really sure what you mean...?

FriedrichFroebel:
You changed `_from_print_page`, `load_quick` and `load_by_guid`. You might add a new test for `_from_print_page` (which does not seem to exist at the moment), while extending `test.test_cache.TestMethods.test_load` and `test.test_cache.TestMethods.test_load_by_guid`.

test/test_cache.py
```diff
@@ -33,7 +33,8 @@ def setUp(self):
             hidden=date(2000, 1, 1),
             attributes={"onehour": True, "kids": False, "available": True},
             summary="text",
-            description="long text",
+            description="engelmz",
+            description_html="<b><font color="red">engelmz</font></b>",
```
FriedrichFroebel, line 37: This is invalid syntax and wrongly repeated twice below. AFAIK the `black` formatter prefers `'<b><font color="red">engelmz</font></b>'` in these cases.

DarkOutcast6: Pretty sure I just fixed it.

DarkOutcast6:
Looks like it's working! 

___

By the way, what does 
image
mean and is it important? 

___

> You changed _from_print_page, load_quick and load_by_guid. You might add a new test for _from_print_page (which does not seem to exist at the moment), while extending test.test_cache.TestMethods.test_load and test.test_cache.TestMethods.test_load_by_guid. 

Do I still need to this? If so, what specifically must I change/add?

FriedrichFroebel (approved):
Thanks for your patience. I have added some small enhancements myself to improve the tests itself.

DarkOutcast6:
Thank you so much for the help!
When will this be released?

DarkOutcast6:
Thank you so much for the help, @FriedrichFroebel! Do you know when this will be added to a release?

FriedrichFroebel:
Version 4.4.0 has been released including this new feature.

PR #224, opened 2023-09-17
Document release process

### LINKED ISSUE #222: Document release process
The current release process is undocumented, especially how to generate the new changelog.

@tomasbedrich Would you be up to document the corresponding steps?

tomasbedrich:
Fixes #222

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -109,3 +109,15 @@ Coding style
   third-party module, please consult it on GitHub before.
 - `Please use regular expressions only as a last resort. <http://imgur.com/j3G9xyP>`_ When possible, use string manipulations,
   such as :code:`split()` and then list operations. It is more readable.
+
+
+Release process
+-------------------------------------------------------------------------------
+
+1. Pick a suitable semantic version number. We adhere to `generic rules <https://docs.npmjs.com/about-semantic-versioning#incrementing-semantic-versions-in-published-packages>`_ with an exception of our `specific deprecation policy <https://github.com/tomasbedrich/pycaching/blob/master/docs/api.rst?plain=1#L7-L11>`_.
+2. If the deprecation policy triggers, remove the deprecated methods. Create a separate PR in that case.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -109,3 +109,15 @@ Coding style
   third-party module, please consult it on GitHub before.
 - `Please use regular expressions only as a last resort. <http://imgur.com/j3G9xyP>`_ When possible, use string manipulations,
   such as :code:`split()` and then list operations. It is more readable.
+
+
+Release process
+-------------------------------------------------------------------------------
+
+1. Pick a suitable semantic version number. We adhere to `generic rules <https://docs.npmjs.com/about-semantic-versioning#incrementing-semantic-versions-in-published-packages>`_ with an exception of our `specific deprecation policy <https://github.com/tomasbedrich/pycaching/blob/master/docs/api.rst?plain=1#L7-L11>`_.
```
tomasbedrich, line 117: @FriedrichFroebel what are your thoughts about the deprecation policy? I am wondering if we should update it to match semver, i.e. for deprecated methods to be removed together with major version bump.

FriedrichFroebel: In my personal opinion, the current deprecation policy is fine and apparently no one complained about it in the past (although deprecations and releases generally are rather sparse). I would reserve major version changes for really breaking changes like dropping old Python versions or old dependency versions, which we do even less.

FriedrichFroebel (approved):

FriedrichFroebel:
@tomasbedrich As the `master` branch is protected, I cannot push to it directly and will have to create a PR which AFAIK usually requires approval from you (or one of the other (currently inactive) collaborators). I do not know if there is a clean solution to this for now, but to clarify this, I did not change the version number for now.

FriedrichFroebel:
I have added a corresponding PR. Yes, the goal is to fulfill the request from #218 while checking that everything works as intended.

FriedrichFroebel:
Thanks, the remaining steps seem to have worked flawlessly.

PR #228, opened 2023-11-28
Allow the string to be used as a `sort_by` parameter in `Geocaching.sort()` method

BelKed:
Allow the string to be used as a `sort_by` parameter in `Geocaching.sort()` method


FriedrichFroebel:
Thanks for the PR. Could you please add a corresponding test as well?

FriedrichFroebel (approved):
LGTM, thanks.

PR #230, opened 2023-12-04
Update RTD configuration file structure

FriedrichFroebel:
While we declared to use the version 2 format for RTD, it seems like it is not valid anymore (or has never been a valid version 2 one).


tomasbedrich (approved):
LGTM, build untested

PR #233, opened 2023-12-27
Fix getting user logs

Jakoma02:
Due to changes on the Geocaching website, iterating `my_logs()` raises an exception due to a missing query argument "guid". This is caused by an upstream change at https://www.geocaching.com/my/logs.aspx, where the format of the URL that this library uses was changed to "https://www.geocaching.com/geocache/[WP]".

This PR modifies the `my_logs()` method to use WP to find the geocache details instead of guid, which fixes the problem.

### DIFF
--- a/pycaching/geocaching.py
+++ b/pycaching/geocaching.py
@@ -8,7 +8,7 @@ import re
 import subprocess
 from os import path
 from typing import Generator, Optional, Union
-from urllib.parse import parse_qs, urljoin, urlparse
+from urllib.parse import urljoin
 
 import bs4
 import requests
@@ -494,8 +494,8 @@ class Geocaching(object):
                 break
 
             link = row.find(class_="ImageLink")["href"]
-            guid = parse_qs(urlparse(link).query)["guid"][0]
-            current_cache = self._try_getting_cache_from_guid(guid)
+            wp = link.split("/")[4]
+            current_cache = self.get_cache(wp)
             date = row.find_all("td")[2].text.strip()
             current_cache.visited = date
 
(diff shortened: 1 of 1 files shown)


pycaching/geocaching.py
```diff
@@ -494,8 +494,8 @@ def my_logs(self, log_type=None, limit=float("inf")):
                 break
 
             link = row.find(class_="ImageLink")["href"]
-            guid = parse_qs(urlparse(link).query)["guid"][0]
-            current_cache = self._try_getting_cache_from_guid(guid)
+            wp = link.split("/")[4]
```
FriedrichFroebel, line 497: Could we make this more explicit somehow or check that the removed prefix is correct? For now, this seems to be "magic", as we do not have automated tests for this at the moment as well.

Jakoma02: An explicit way of doing this might be using regular expressions, which are not recommended as per contribution guidelines. Also note that on [line 470](https://github.com/tomasbedrich/pycaching/blob/15bc2acf2ba6e9141e850fb8296d8cf5058cce7f/pycaching/geocaching.py#L470), a very similar extraction is done in a very similar way.

Would you consider this resolved if a comment is provided similarly to line 470? Or would you rather see explicit checking of the prefix or moving the "logic" to a separate function?

FriedrichFroebel: In the easiest way, we can consider adding a comment with an example URL here to make this more clear.

Jakoma02: I added the comment in commit cdeeeca, please see if this is sufficient.

FriedrichFroebel (approved):
Thanks, LGTM.

ohheyrj:
Has this been released yet or is it only in master at the moment?

tomasbedrich:
Now it is released as 4.4.1

PR #235, opened 2024-02-08
Add support for archive log type "6"

GeoTime61:
Most caches use an archive log type of "5", but some caches use an archive type of "6". Retain support for type "5" and add support for type "6".

### DIFF
--- a/pycaching/log.py
+++ b/pycaching/log.py
@@ -143,6 +143,9 @@ class Type(enum.Enum):
         elif filename == "68":
             # 2 different IDs for post_reviewer_note
             return cls.post_reviewer_note
+        elif filename == "6":
+            # 2 different IDs for archive
+            return cls.archive
         elif filename == "1":
             # 2 different IDs for unarchive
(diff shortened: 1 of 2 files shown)


GeoTime61:
I don't know anything about GitHub, but I did my best to create a pull request with the change and a simple test.

Following the steps in CONTRIBUTING:
"isort ." worked.
"black ." worked.
"flake8" failed - AttributeError: 'EntryPoints' object has no attribute 'get'
"pytest test" (had to add " test") failed - FAIL Required test coverage of 90% not reached. Total coverage: 89.90%

Maybe you can tell me what I did wrong or help me fix the errors.

FriedrichFroebel:
Thanks for the PR.

Running `python -m pip install .[dev]` before running *flake8* or *pytest* should in theory not yield any issues and work out of the box - different versions might yield different results. The GitHub Actions CI is using the same commands without any real issues.

FriedrichFroebel (approved):
LGTM.

PR #249, opened 2026-09-16
Fix attributes and inventory widgets

joelhoffman:
I discovered that attributes were not being parsed from the cache page correctly due to a change in the order of div elements at some point. This should fix it. I'm not sure the inventory widget here gets used directly but I think it should be pointing to the analogous div it pointed to previously.


pycaching/cache.py
```diff
@@ -765,7 +765,13 @@ def load(self):
         pm_only_warning = root.find("p", "Warning NoBottomSpacing")
         self.pm_only = pm_only_warning and ("Premium Member Only" in pm_only_warning.text) or False
 
-        attributes_widget, inventory_widget, *_ = root.find_all("div")
+        inventory_widget = None
+        attributes_widget = None
+        for header in root.find_all("h3","WidgetHeader"):
+            if header.text.strip() == "Attributes":
```
FriedrichFroebel, line 771: I somehow remember that we would try to avoid issues with different languages. Is this the only reliable way to determine this (and thus require an English locale), or are there language-independent alternatives?

test/test_cache.py
```diff
@@ -343,6 +343,27 @@ def test_load_trackables(self):
             trackable_list = list(cache.load_trackables(limit=10))
         self.assertTrue(isinstance(trackable_list, list))
 
+
+    def test_load_attributes(self):
```
FriedrichFroebel, line 347: Between methods, there usually should be exactly one empty line.

FriedrichFroebel (changes requested):
Thanks for the PR. Besides the inline feedback, could you please check the CI failures?

pycaching/cache.py
joelhoffman: I did look for alternatives, but maybe not carefully enough. Will revise, I'm sure there's another alternative.

